PR #19, opened 2022-07-28
✨ Add divUp in Math lib

MerlinEgalite:


src/math/Math.sol
```diff
@@ -28,4 +28,16 @@ library Math {
             }
         }
     }
+
+    /// @dev Returns x / y rounded up.
+    function divUp(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        assembly {
+            // Revert if a + b < a or a + b < b
```
Rubilmax, line 35:
```suggestion
            // Overflows if a + b < a or a + b < b
```

Rubilmax: and thus it explains why we revert (I'd move this comment down a line ideally)

Rubilmax (approved):

src/math/Math.sol
```diff
@@ -28,4 +28,16 @@ library Math {
...
+            z := add(x, y)
+            if or(iszero(y), or(lt(z, y), lt(z, x))) {
+                revert(0, 0)
+            }
+            z := div(sub(z, 1), y)
```
ewile, line 40: Save 9 gas.
(Need to define at the beginning: `uint256 internal constant MAX_UINT256 = 2**256 - 1;`)

```suggestion
            z := add(x, y)
            if or(iszero(y), or(lt(z, y), lt(z, x))) {
                revert(0, 0)
            }
            z := div(sub(z, 1), y)
```


```suggestion
            if or(iszero(y), gt(x, sub(MAX_UINT256, y))) {
                revert(0, 0)
            }
            z := div(sub(add(x, y), 1), y)
```
image

ewile (dismissed):

src/math/Math.sol
pakim249CAL: The only way a + b < a or b is if  a + b overflowed, right? In that case, we should only need to do one check against a or b. and then we can change the comment to something like:
// Reverts if a + b overflows

pakim249CAL: Proof:

Let n = 2^256. We want to check that a + b < n for overflow.
Let x = (a + b) % n
a < n & b < n

If a + b < n, then x = a + b
x >= a

If a + b >= n, then x = a + b - n
b < n so
x < a + n - n = a

This math works for a or b.

MerlinEgalite: Good point @pakim249CAL

MerlinEgalite: agree with you too @Rubilmax

src/math/Math.sol
MerlinEgalite:
```solidity
assembly {
            z := add(x, y)
            if or(iszero(y), lt(z, x)) {
                revert(0, 0)
            }
            z := div(sub(z, 1), y)
}```

This version is even better -> 448

File: src/math/Math.sol (hunk 1 above)
Comment (MerlinEgalite) [reply src/math/Math.sol:35]: Tried @ewile version but this one seems to be better.

**Review by pakim249CAL — APPROVED** [review]

File: src/math/Math.sol [hunk 3]
```diff
@@ -28,4 +28,16 @@ library Math {
...
+    function divUp(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        assembly {
+            z := add(x, y)
+            // Revert if y = 0 or if x + y < x
+            if or(iszero(y), lt(z, x)) {
+                revert(0, 0)
+            }
+            z := div(sub(z, 1), y)
+        }
+    }
```
AtomicAzzaz, line 42: Currently, the MathRef divUp reverts even on theoretically possible 256 bits operations like divUp(MAX_UINT256, 10).
If we want to extend the definition of divUp to every operation possible - reverting only on division by 0 - it is possible and cheaper.
The question is therefore if we want to replicate Aave's behaviour or create a more general function.
Tests must be modified (expect revert only when y = 0 / impossible to compare to MathRef on new scenarios).
```suggestion
    /// @dev Returns x / y rounded up.
    function divUp(uint256 x, uint256 y) internal pure returns (uint256 z) {
        assembly {
            // Revert if y = 0
            if iszero(y){
                revert(0,0)
            }
                
            if gt(mod(x, y), 0) {
                z := 1
            }
            z := add(z,div(x, y))

        }
    }
```
We must demonstrate that z doesn't overflow. 
If x % y = 0, z = 0 + x/y <= x <= uint_max. 
Else, x % y != 0 so x = ky + p with 0 < p < y. z= 1 + x//y = 1 + k overflow only if k >= UINT_MAX. 
But if k >= UINT_MAX, p > 0, y > 0 => x = ky + p > UINT_MAX which is not possible.

Gas reports show use of 444 gas when x is a multiple of y and 449 when it's not. 
We can assume that in the general case, y doesn't divide x and so previous commit (448) might be slightly better in the general case.
Avg gas cost over x knowing y is indeed only better when y < 6
Let's try to negate the second condition: 
```suggestion
    /// @dev Returns x / y rounded up.
    function divUp(uint256 x, uint256 y) internal pure returns (uint256 z) {
        assembly {
            // Revert if y = 0
            if iszero(y){
                revert(0,0)
            }
        
            z := 1
            if iszero(mod(x, y)) {
                z := 0
            }

            z := add(z,div(x, y))

        }
    }
```
Obviously, the demonstration still holds.
Gas reports show use of 451 (+7) gas when x is a multiple of y and 446 (-3) when it's not.
Avg gas cost over x knowing y is better than previous commit as soon as y > 2.
Again, overall behaviour changes here which might be unwanted.

MerlinEgalite: This function is not part of Aave but our code, so it's more our choice. Not sure it's safe to do it.

MathisGD: I agree with @AtomicAzzaz

QGarchery: Wait those functions do not behave the same as the original function, even when there is no overflow. For `x == 0` this returns `z >= 1`. This introduces could introduce small borrows for someone who is a pure suppliers, making the complexity of the future functions significantly higher

test/TestMath.sol
```diff
@@ -56,6 +65,17 @@ contract TestMath is Test {
...
+        unchecked {
+            if (y == 0 || (x + y < x)) {
+                vm.expectRevert();
+                Math.divUp(x, y);
+            }
+        }
+
+        assertEq(math.divUp(x, y), mathRef.divUp(x, y));
+    }
+
```
QGarchery, line 78: Test `divUp` when there is no remainder

QGarchery (approved):

src/math/Math.sol
QGarchery: Wait those functions do not behave the same as the original function, even when there is no overflow. For x == 0 this returns z >= 1. This introduces could introduce small borrows for someone who is a pure suppliers, making the complexity of the future functions significantly higher

QGarchery: My bad, this is fine actually. Let's do it !

src/math/Math.sol
```diff
@@ -28,4 +28,15 @@ library Math {
             }
         }
     }
+
+    /// @dev Returns x / y rounded up ((x + y - 1) / y).
```
AtomicAzzaz, line 32:
```suggestion
    /// @dev Returns x / y rounded up ( x/y + boolAsInt(x%y>0) ).
```

AtomicAzzaz (approved):

test/TestMath.sol
```diff
@@ -56,6 +65,19 @@ contract TestMath is Test {
...
+            if (y == 0) {
+                vm.expectRevert();
+                Math.divUp(x, y);
+            } else if (x + y < x) {
+                assertApproxEqAbs(math.divUp(x, y), x / y, 1);
+            } else {
+                assertEq(math.divUp(x, y), mathRef.divUp(x, y));
+            }
+        }
+    }
```
Rubilmax, line 79:
```suggestion
    function testDivUpRevertWhenDivByZero(uint256 x) public {
        vm.expectRevert();
        Math.divUp(x, 0);
    }

    function testDivUpWhenNumSmaller(uint256 x, uint256 y) public {
        vm.assume(x < y);
        assertEq(math.divUp(x, y), 1);
    }

    function testDivUpWhenOperandsEqual(uint256 x) public {
        assertEq(math.divUp(x, x), 1);
    }

    function testDivUpWhenNumLargerAndDivisible(uint256 x, uint256 y) public {
        vm.assume(x > y);
        vm.assume(x % y == 0);
        assertEq(math.divUp(x, y), x / y);
    }

    function testDivUpWhenNumLargerAndNotDivisible(uint256 x, uint256 y) public {
        vm.assume(x > y);
        vm.assume(x % y != 0);
        assertEq(math.divUp(x, y), x / y + 1);
    }
```

Rubilmax (approved):

Rubilmax (commented):
oups

PR #25, opened 2022-07-29
Add a delegateCall test regarding memory

AtomicAzzaz:
This test asserts that the delegateCall function of the delegateCall library leaves the memory in the same state as a the targetDelegateCallBehaviour function - which was the first implementation of the delegateCall library before being optimized in yul. This test compares the free memory pointer value and the full memory state using both functions. 

The idea behind targetDelegateCallBehaviourAndReturnMemoryAndPointer is to write in the memory before and after the delegateCall to not be in a trivial state with only 0's in memory. We test the memory on various data types returned by the delegateCall. 

Note: it is possible to make a full working implementation having this test failing (like having the free mem pointer further than ref implem) but is I believe not safe to do.

### DIFF
--- a/test/TestDelegateCall.sol
+++ b/test/TestDelegateCall.sol
@@ -27,7 +27,7 @@ contract Caller {
         bytes32 ptr;
         assembly{
             ptr := mload(0x40)
-            //We write a slot in the memory before the call 
+            // We write a slot in the memory before the call.
             mstore(ptr, 0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb)
             mstore(0x40, add(ptr, 0x20))
         }
@@ -36,11 +36,11 @@ contract Caller {
         
         assembly{
             ptr := mload(0x40)
-            //We write a slot in the memory after the call
+            // We write a slot in the memory after the call.
             mstore(ptr, 0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa)
             mstore(0x40, add(ptr, 0x20))
         }
-        //We store the memory in allMem
+        // We store the memory in allMem.
         assembly{
             ptr := mload(0x40)
             allMem := 0x00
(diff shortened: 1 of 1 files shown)


Rubilmax (approved):
useful!
Could we extract `targetDelegateCallBehaviour` to a DelegateCallRef contract?
To keep the same architecture as for the other utils

test/TestDelegateCall.sol
```diff
@@ -20,11 +20,40 @@ contract Caller {
...
         return abi.decode(data, (uint256));
     }
+
+
+    function delegateCallAndReturnMemoryAndPointer(bytes4 _selector) external returns (bytes memory, bytes32) {
+        bytes memory allMem;
+        bytes32 ptr;
+        assembly{
+            ptr := mload(0x40)
+            //We write a slot in the memory before the call
```
MerlinEgalite, line 30:
```suggestion
            // We write a slot in the memory before the call.
```

test/TestDelegateCall.sol
```diff
@@ -20,11 +20,40 @@ contract Caller {
...
+            //We write a slot in the memory before the call 
+            mstore(ptr, 0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb)
+            mstore(0x40, add(ptr, 0x20))
+        }
+
+        called.functionDelegateCall(abi.encodeWithSelector(_selector, ""));
+        
+        assembly{
+            ptr := mload(0x40)
+            //We write a slot in the memory after the call
```
MerlinEgalite, line 39:
```suggestion
            // We write a slot in the memory after the call.
```

test/TestDelegateCall.sol
```diff
@@ -20,11 +20,40 @@ contract Caller {
...
+            mstore(ptr, 0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa)
+            mstore(0x40, add(ptr, 0x20))
+        }
+        //We store the memory in allMem
```
MerlinEgalite, line 43:
```suggestion
        // We store the memory in allMem.
```

test/TestDelegateCall.sol
```diff
@@ -92,6 +174,35 @@ contract TestDelegateCall is Test {
         caller.delegateCall(Called.revertWithCustomError.selector);
     }
 
+    function testMemoryUncorruptedWithRef() public{
+        //This test makes sure that the delegateCall function of our library acts on the memory in the exact same way 
+        //as a targetBehaviour function corresponding to the desired behavior.
```
MerlinEgalite, line 179:
```suggestion
        // This test makes sure that the delegateCall function of our library acts on the memory in the exact same way 
        // as a targetBehaviour function corresponding to the desired behavior.
```

MerlinEgalite (changes requested):
Small changes to apply but overall LGTM

MerlinEgalite (approved):

PR #29, opened 2022-07-30
Improve testing

### LINKED ISSUE #27: Extract fuzzing test branches to separate, pre-assumed fuzzing test
Basically extract every:

```solidity
function test(uint256 blabla) {
    if (condition) {
        expect(expectation1);
    } else {
        expect(expectation2);
    }
}
```

to:

```solidity
function testWhenCondition(uint256 blabla) {
    vm.assume(condition);
    expect(expectation1);
}

function testWhenNotCondition(uint256 blabla) {
    vm.assume(!condition);
    expect(expectation2);
}
```

MathisGD:
Fixes #27 and small refactoring.


test/TestPercentageMath.sol
```diff
@@ -43,27 +43,38 @@ contract TestPercentageMath is Test {
     /// TESTS ///
 
     function testPercentMul(uint256 x, uint256 y) public {
-        unchecked {
-            if (y > 0 && x > MAX_UINT256_MINUS_HALF_PERCENTAGE / y) {
-                vm.expectRevert();
-                PercentageMath.percentMul(x, y);
-            }
-        }
+        vm.assume(y == 0 || (y > 0 && x <= MAX_UINT256_MINUS_HALF_PERCENTAGE / y));
```
Rubilmax, line 46:
```suggestion
        vm.assume(y == 0 || x <= MAX_UINT256_MINUS_HALF_PERCENTAGE / y);
```

Rubilmax (approved):
much, much cleaner, thank you!

Rubilmax:
Note that extracting source code branches into several, separate fuzzing tests does not impact gas cost (neither its estimation, nor its "true" value)
Could we thus infer that the fuzzing engine is optimized to browser every branch of the source code?

I still prefer separating fuzzing tests into separate, clear test cases though

.prettierrc.json
```diff
@@ -11,4 +10,4 @@
       }
     }
   ]
-}
+}
```
QGarchery, line 13: I like newlines at end of files :)
(it's a unix convention, it makes text formatting easier for some tools)

test/TestWadRayMath.sol
```diff
@@ -60,14 +60,14 @@ contract WadRayMathFunctionsRef {
...
-    uint256 public constant HALF_RAY = 0.5e27;
+    uint256 public constant HALF_RAY = RAY / 2;
     uint256 public constant WAD_RAY_RATIO = 1e9;
-    uint256 public constant HALF_WAD_RAY_RATIO = 0.5e9;
+    uint256 public constant HALF_WAD_RAY_RATIO = WAD_RAY_RATIO / 2;
     uint256 public constant MAX_UINT256 = 2**256 - 1; // Not possible to use type(uint256).max in yul.
-    uint256 public constant MAX_UINT256_MINUS_HALF_WAD = 2**256 - 1 - 0.5e18;
-    uint256 public constant MAX_UINT256_MINUS_HALF_RAY = 2**256 - 1 - 0.5e27;
+    uint256 public constant MAX_UINT256_MINUS_HALF_WAD = MAX_UINT256 - HALF_WAD;
+    uint256 public constant MAX_UINT256_MINUS_HALF_RAY = MAX_UINT256 - HALF_RAY;
```
QGarchery, line 70: In tests we can also use `type(uint256).max` for `MAX_UINT256`

test/TestCompoundMath.sol
```diff
@@ -67,27 +87,7 @@ contract TestCompoundMath is Test {
...
-
-contract CompoundMathFunctionsRef {
-    function mul(uint256 x, uint256 y) public pure returns (uint256) {
-        return CompoundMathRef.mul(x, y);
-    }
-
-    function div(uint256 x, uint256 y) public pure returns (uint256) {
-        return CompoundMathRef.div(x, y);
+        compoundMath.div(2 * WAD, WAD);
+        compoundMathRef.div(2 * WAD, WAD);
```
QGarchery, line 91: There are some lower case `compoundmath` left. We can also remove those variables and the `setUp` function

QGarchery (approved):
Looks good, a few small improvements can be made

.prettierrc.json
Rubilmax: There's also a setting dedicated to this in vscode: https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files

QGarchery:
Not completely related but it would be nice if you could also add the `forge-std` remappings

test/TestCompoundMath.sol
Rubilmax: Nope, these are used for gas report and need a contract rather than a library (because Foundry does not generate gas reports of libraries)

QGarchery: Oh ok, my bad

test/TestWadRayMath.sol
MerlinEgalite: not sure it's good to change that as it's thus different from libs we are testing which is weird...

MerlinEgalite (approved):
Just added a comment following @QGarchery's comment, not sure we should change it but as you want

.prettierrc.json
MathisGD: Hmm, weird that I have to change settings for this.. We may not use the same formatter ? 🤔

test/TestWadRayMath.sol
MathisGD: I'm in favor of changing it too.

test/TestPercentageMath.sol
MathisGD: Not sure that this is possible (div by 0)

MathisGD:
> Note that extracting source code branches into several, separate fuzzing tests does not impact gas cost (neither its estimation, nor its "true" value) Could we thus infer that the fuzzing engine is optimized to browse every branch of the source code?
> 
> I still prefer separating fuzzing tests into separate, clear test cases though

I also have mixed feeling about this.. Either fuzzing at a higher level and fuzzing each branch have their advantages.

test/TestPercentageMath.sol
MathisGD: oh no you are right, if the first assertion is true, it doesn't evaluate the second one in an OR.

QGarchery (approved):

PR #31, opened 2022-08-01
⚡️ Opti conditions

makcandrov:
`percentageMul`: 413 => 409
`percentageDiv`: 403 => 402

`rayMul`: 479 => 475
`rayDiv`: 469 => 468
`wadMul`: 435 => 431
`wadDiv` 513 => 512

### DIFF
--- a/src/math/WadRayMath.sol
+++ b/src/math/WadRayMath.sol
@@ -30,7 +30,7 @@ library WadRayMath {
         // <=> x * y > type(uint256).max - HALF_WAD
         // <=> x > (type(uint256).max - HALF_WAD) / y
         assembly {
-            if and(gt(y, 0), gt(x, div(MAX_UINT256_MINUS_HALF_WAD, y))) {
+            if mul(y, gt(x, div(MAX_UINT256_MINUS_HALF_WAD, y))) {
                 revert(0, 0)
             }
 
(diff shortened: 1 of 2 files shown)


AtomicAzzaz:
Nice one, works also in wadRayMath for 4 gas savings in differents Mul

Rubilmax (approved):
jpp
it gets unreadable but it works

makcandrov:
> Nice one, works also in wadRayMath for 4 gas savings in differents Mul

Updated in the last commit :)

src/math/WadRayMath.sol
```diff
@@ -30,7 +30,7 @@ library WadRayMath {
         // <=> x * y > type(uint256).max - HALF_WAD
         // <=> x > (type(uint256).max - HALF_WAD) / y
         assembly {
-            if and(gt(y, 0), gt(x, div(MAX_UINT256_MINUS_HALF_WAD, y))) {
+            if mul(y, gt(x, div(MAX_UINT256_MINUS_HALF_WAD, y))) {
```
QGarchery, line 33: To be precise we can add the condition `y != 0` in the last line in the comment above

QGarchery (approved):
Ok this should work yes. Although I don't understand why the div opti saves some gas: replace one OR (3) with both a MUL (5) and a ISZERO (3) ...

src/math/WadRayMath.sol
makcandrov: `y` is already declared positive L28

```solidity
// Let y > 0
```

QGarchery: True, I missed that. But in the case of multiplication it's not clear why we would assume that at the start. In contrast, the last equivalence makes it clear why we are checking this condition: we are doing a division by `y`

MerlinEgalite (approved):
LGTM

MathisGD (commented):
LGTM

PR #33, opened 2022-08-02
Opti CompoundMath

makcandrov:
before:
image

after:
image


Rubilmax (approved):

QGarchery (approved):
Looks good ! The optimization of the div function is of 6 gas because the compiler decides to PUSH 0, SWAP, POP in the original implementation, and not in the implementation of this PR. It thus saves 3 + 3 + 2 (POP) and loses 2 (MUL instead of OR). I don't know why it did that though

MerlinEgalite (approved):
LGTM

PR #36, opened 2022-08-03
Fix too strong conditions in tests

MathisGD:


test/TestMath.sol
```diff
@@ -83,14 +83,12 @@ contract TestMath is Test {
 
     function testDivUpWhenNumLargerAndDivisible(uint256 x, uint256 y) public {
         vm.assume(y > 0);
-        vm.assume(x > y);
```
QGarchery, line 86: We can now change the name of the function

test/TestMath.sol
```diff
@@ -83,14 +83,12 @@ contract TestMath is Test {
...
         vm.assume(x % y == 0);
         assertEq(Math.divUp(x, y), x / y);
     }
 
     function testDivUpWhenNumLargerAndNotDivisible(uint256 x, uint256 y) public {
         vm.assume(y > 0);
```
QGarchery, line 91: same

QGarchery (approved):

test/TestMath.sol
```diff
@@ -70,27 +70,25 @@ contract TestMath is Test {
...
 
-    function testDivUpWhenNumLargerAndDivisible(uint256 x, uint256 y) public {
+    function testDivUpDivisible(uint256 x, uint256 y) public {
         vm.assume(y > 0);
-        vm.assume(x > y);
         vm.assume(x % y == 0);
         assertEq(Math.divUp(x, y), x / y);
     }
 
-    function testDivUpWhenNumLargerAndNotDivisible(uint256 x, uint256 y) public {
```
Rubilmax, line 91: I know test names are large but it makes reading a test report & the code easier by giving a clear context
I don't think it's the case for `testDivUpNotDivisible` which does not indicate whether it's the numerator or the denominator which is not divisible

I would recommend `testDivUpNumNotDivisible` and `testDivUpNumDivisible`, if making test names shorter is meaningful

Rubilmax (approved):

QGarchery (approved):

MerlinEgalite (approved):
LGTM

test/TestMath.sol
MathisGD: >which does not indicate whether it's the numerator or the denominator which is not divisible
this is implicit

But I can re-add that if you want!

MathisGD: >which does not indicate whether it's the numerator or the denominator which is not divisible
this is implicit

But I can re-add that if you want!

PR #38, opened 2022-08-05
perf: opti mul args order

MathisGD:
🏌️

### DIFF
--- a/src/math/CompoundMath.sol
+++ b/src/math/CompoundMath.sol
@@ -34,7 +34,7 @@ library CompoundMath {
                 revert(0, 0)
             }
 
-            z := div(div(mul(x, SCALE), y), WAD)
+            z := div(div(mul(SCALE, x), y), WAD)
         }
     }
 }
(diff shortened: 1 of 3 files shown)


pakim249CAL (approved):

Rubilmax (approved):

AtomicAzzaz (approved):

MerlinEgalite (approved):
LGTM

QGarchery (approved):
Quick analysis of this opti.
Let's call situation 1 when you compute `mul(x, SCALE)` and situation 2 when you compute `mul(SCALE, x)` (like after the opti). In both cases the stack is initially `SCALE WAD y x`  where the top of the stack is `SCALE`. We want to arrive at a stack where we can do  `MUL DIV DIV`. Both are achievable in 2 swaps with `SWAP1 SWAP3`, which is what is found by the compiler in situation 2, but the compiler finds only how to do it in 3 swaps in situation 1 with `SWAP3 SWAP1 SWAP3`

PR #40, opened 2022-08-11
Perf percentage math

MerlinEgalite:
We could improve the average computation if we do the following:
`(x * (1 - percentage) + y * percentage) / PERCENTAGE_FACTOR` instead of:
`x * (1 - percentage)  / PERCENTAGE_FACTOR + y * percentage / PERCENTAGE_FACTOR`

Note that it's not the same computation with the roundings and also it would overflow with a lower (x, y) pair.

### DIFF
--- a/src/math/PercentageMath.sol
+++ b/src/math/PercentageMath.sol
@@ -66,10 +66,9 @@ library PercentageMath {
     /// @param percentage The percentage of the value to multiply.
     /// @return y The result of the multiplication.
     function percentMul(uint256 x, uint256 percentage) internal pure returns (uint256 y) {
-        // Let percentage > 0
-        // Overflow if x * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
-        // <=> x * percentage > type(uint256).max - HALF_PERCENTAGE_FACTOR
-        // <=> x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
+        // Should revert if
+        // x * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        // <=> percentage > 0 and x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
         assembly {
             if mul(percentage, gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, percentage))) {
                 revert(0, 0)
@@ -84,12 +83,13 @@ library PercentageMath {
     /// @param percentage The percentage of the value to divide.
     /// @return y The result of the division.
     function percentDiv(uint256 x, uint256 percentage) internal pure returns (uint256 y) {
-        // let percentage > 0
-        // Overflow if x * PERCENTAGE_FACTOR + halfPercentage > type(uint256).max
-        // <=> x * PERCENTAGE_FACTOR > type(uint256).max - halfPercentage
-        // <=> x > type(uint256).max - halfPercentage / PERCENTAGE_FACTOR
+        // Should revert if
+        // percentage == 0
+        // or x * PERCENTAGE_FACTOR + percentage / 2 > type(uint256).max
+        // <=> percentage == 0
+        // or x > (type(uint256).max - percentage / 2) / PERCENTAGE_FACTOR
         assembly {
-            y := div(percentage, 2) // Temporary assignement to save gas.
+            y := div(percentage, 2) // Temporary assignment to save gas.
 
             if iszero(mul(percentage, iszero(gt(x, div(sub(MAX_UINT256, y), PERCENTAGE_FACTOR))))) {
                 revert(0, 0)
--- a/src/math/PercentageMath.sol
+++ b/src/math/PercentageMath.sol
@@ -53,7 +53,7 @@ library PercentageMath {
         assembly {
             y := sub(PERCENTAGE_FACTOR, percentage) // Temporary assignment to save gas.
 
-            if or(gt(percentage, PERCENTAGE_FACTOR), and(y, gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, y)))) {
+            if or(gt(percentage, PERCENTAGE_FACTOR), mul(y, gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, y)))) {
                 revert(0, 0)
             }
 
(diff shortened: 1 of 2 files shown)


Rubilmax:
> We could improve the average computation if we do the following:
> `(x * (1 - percentage) + y * percentage) / PERCENTAGE_FACTOR` instead of:
> `x * (1 - percentage)  / PERCENTAGE_FACTOR + y * percentage / PERCENTAGE_FACTOR`
> 
> Note that it's not the same computation with the roundings and also it would overflow with a lower (x, y) pair.

Definitely to avoid because of the 2 points you made

Will check this PR out, thanks!

src/math/PercentageMath.sol
```diff
@@ -72,12 +104,41 @@ library PercentageMath {
...
+        // Let percentage > 0
+        // Underflow if percentage > PERCENTAGE_FACTOR
+        //
+        // Overflow if x1 = x * (PERCENTAGE_FACTOR - percentage) + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        // <=> x * (PERCENTAGE_FACTOR - percentage) > type(uint256).max - HALF_PERCENTAGE_FACTOR
+        // <=> x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR - percentage)
+        //
+        // Overflow if x2 = y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        // <=> y * percentage > type(uint256).max - HALF_PERCENTAGE_FACTOR
+        // <=> y > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
```
MathisGD, line 122: percentage should be greater than 0

MathisGD (commented):
I'm opening a PR on top of this one to fix all of this

src/math/PercentageMath.sol
MerlinEgalite: let percentage > 0

src/math/PercentageMath.sol
```diff
@@ -18,17 +18,48 @@ library PercentageMath {
...
+        // <=> percentage > type(uint256).max - PERCENTAGE_FACTOR or x * (PERCENTAGE_FACTOR + percentage) > type(uint256).max - HALF_PERCENTAGE_FACTOR
+        // <=> percentage > type(uint256).max - PERCENTAGE_FACTOR or x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR + percentage)
+        assembly {
+            y := add(percentage, PERCENTAGE_FACTOR) // Temporary assignement to save gas.
+
+            if mul(
+                y,
+                or(
+                    gt(percentage, sub(MAX_UINT256, PERCENTAGE_FACTOR)),
+                    gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, y))
```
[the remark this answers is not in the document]
MerlinEgalite: wondering the same thing, we know that operations between constants is optimized after compilation, so either we only use the operations or the long constants even we know that it does not bring any gas optimizations.

src/math/PercentageMath.sol
MathisGD: yes, but it is weird this way IMO

src/math/PercentageMath.sol
```diff
@@ -18,17 +18,48 @@ library PercentageMath {
...
+    function percentAdd(uint256 x, uint256 percentage) internal pure returns (uint256 y) {
+        // Let percentage > 0
+        // Overflow if PERCENTAGE_FACTOR + percentage > type(uint256).max or x * (PERCENTAGE_FACTOR + percentage) + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        // <=> percentage > type(uint256).max - PERCENTAGE_FACTOR or x * (PERCENTAGE_FACTOR + percentage) > type(uint256).max - HALF_PERCENTAGE_FACTOR
+        // <=> percentage > type(uint256).max - PERCENTAGE_FACTOR or x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR + percentage)
+        assembly {
+            y := add(percentage, PERCENTAGE_FACTOR) // Temporary assignement to save gas.
+
+            if mul(
+                y,
```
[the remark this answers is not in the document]
MerlinEgalite: not sure to understand sorry

src/math/PercentageMath.sol
```diff
@@ -72,12 +104,41 @@ library PercentageMath {
...
+
+            if or(
+                or(
+                    mul(
+                        sub_,
+                        or(gt(percentage, PERCENTAGE_FACTOR), gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, sub_)))
+                    ),
+                    mul(percentage, gt(y, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, percentage)))
+                ),
+                lt(z, x1)
```
MathisGD, line 138: x1 + x2 is always less than MAX
knowing this, we can optimize a lot this one

src/math/PercentageMath.sol
```diff
@@ -72,12 +104,41 @@ library PercentageMath {
...
         uint256 percentage
-    ) internal pure returns (uint256) {
-        return percentMul(x, PERCENTAGE_FACTOR - percentage) + percentMul(y, percentage);
+    ) internal pure returns (uint256 z) {
+        // Let percentage > 0
+        // Underflow if percentage > PERCENTAGE_FACTOR
+        //
+        // Overflow if x1 = x * (PERCENTAGE_FACTOR - percentage) + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        // <=> x * (PERCENTAGE_FACTOR - percentage) > type(uint256).max - HALF_PERCENTAGE_FACTOR
+        // <=> x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR - percentage)
```
MathisGD, line 118: `(PERCENTAGE_FACTOR - percentage)` should be greater than 0

MerlinEgalite: first condition if not passed then it reverts

src/math/PercentageMath.sol
MerlinEgalite: hmm why?

src/math/PercentageMath.sol
MathisGD: Not exactly 
There is the case percentage = PERCENTAGE_FACTOR

src/math/PercentageMath.sol
MathisGD: Because x1 + x2 <= max(x, y)

src/math/PercentageMath.sol
MerlinEgalite: Oh ok understand now what you mean

src/math/PercentageMath.sol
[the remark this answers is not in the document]
MathisGD: percentage <= MAX - PERCENTAGE_FACTOR => percentage + PERCENTAGE_FACTOR < MAX + 1 => y != 0

so no need to verify that y != 0

PR #41, opened 2022-08-12
Fixes proposal after review

MathisGD:
Improvements on top of #40. 
Better reviewed commit by commit.

### DIFF
--- a/src/math/PercentageMath.sol
+++ b/src/math/PercentageMath.sol
@@ -66,10 +66,9 @@ library PercentageMath {
     /// @param percentage The percentage of the value to multiply.
     /// @return y The result of the multiplication.
     function percentMul(uint256 x, uint256 percentage) internal pure returns (uint256 y) {
-        // Let percentage > 0
-        // Overflow if x * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
-        // <=> x * percentage > type(uint256).max - HALF_PERCENTAGE_FACTOR
-        // <=> x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
+        // Should revert if
+        // x * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        // <=> percentage > 0 and x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
         assembly {
             if mul(percentage, gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, percentage))) {
                 revert(0, 0)
@@ -84,12 +83,13 @@ library PercentageMath {
     /// @param percentage The percentage of the value to divide.
     /// @return y The result of the division.
     function percentDiv(uint256 x, uint256 percentage) internal pure returns (uint256 y) {
-        // let percentage > 0
-        // Overflow if x * PERCENTAGE_FACTOR + halfPercentage > type(uint256).max
-        // <=> x * PERCENTAGE_FACTOR > type(uint256).max - halfPercentage
-        // <=> x > type(uint256).max - halfPercentage / PERCENTAGE_FACTOR
+        // Should revert if
+        // percentage == 0
+        // or x * PERCENTAGE_FACTOR + percentage / 2 > type(uint256).max
+        // <=> percentage == 0
+        // or x > (type(uint256).max - percentage / 2) / PERCENTAGE_FACTOR
         assembly {
-            y := div(percentage, 2) // Temporary assignement to save gas.
+            y := div(percentage, 2) // Temporary assignment to save gas.
 
             if iszero(mul(percentage, iszero(gt(x, div(sub(MAX_UINT256, y), PERCENTAGE_FACTOR))))) {
                 revert(0, 0)
--- a/src/math/PercentageMath.sol
+++ b/src/math/PercentageMath.sol
@@ -53,7 +53,7 @@ library PercentageMath {
         assembly {
             y := sub(PERCENTAGE_FACTOR, percentage) // Temporary assignment to save gas.
 
-            if or(gt(percentage, PERCENTAGE_FACTOR), and(y, gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, y)))) {
+            if or(gt(percentage, PERCENTAGE_FACTOR), mul(y, gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, y)))) {
                 revert(0, 0)
             }
 
--- a/src/math/PercentageMath.sol
+++ b/src/math/PercentageMath.sol
@@ -22,9 +22,9 @@ library PercentageMath {
     function percentAdd(uint256 x, uint256 percentage) internal pure returns (uint256 y) {
         // Must revert if
         // PERCENTAGE_FACTOR + percentage > type(uint256).max
-        // or x * (PERCENTAGE_FACTOR + percentage) + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        //     or x * (PERCENTAGE_FACTOR + percentage) + HALF_PERCENTAGE_FACTOR > type(uint256).max
         // <=> percentage > type(uint256).max - PERCENTAGE_FACTOR
-        // or x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR + percentage)
+        //     or x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR + percentage)
         // Note: PERCENTAGE_FACTOR + percentage >= PERCENTAGE_FACTOR > 0
         assembly {
             y := add(PERCENTAGE_FACTOR, percentage) // Temporary assignment to save gas.
@@ -47,9 +47,9 @@ library PercentageMath {
     function percentSub(uint256 x, uint256 percentage) internal pure returns (uint256 y) {
         // Must revert if
         // percentage > PERCENTAGE_FACTOR
-        // or x * (PERCENTAGE_FACTOR - percentage) + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        //     or x * (PERCENTAGE_FACTOR - percentage) + HALF_PERCENTAGE_FACTOR > type(uint256).max
         // <=> percentage > PERCENTAGE_FACTOR
-        // or ((PERCENTAGE_FACTOR - percentage) > 0 and x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR - percentage))
+        //     or ((PERCENTAGE_FACTOR - percentage) > 0 and x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR - percentage))
         assembly {
             y := sub(PERCENTAGE_FACTOR, percentage) // Temporary assignment to save gas.
 
@@ -85,9 +85,9 @@ library PercentageMath {
     function percentDiv(uint256 x, uint256 percentage) internal pure returns (uint256 y) {
         // Must revert if
         // percentage == 0
-        // or x * PERCENTAGE_FACTOR + percentage / 2 > type(uint256).max
+        //     or x * PERCENTAGE_FACTOR + percentage / 2 > type(uint256).max
         // <=> percentage == 0
-        // or x > (type(uint256).max - percentage / 2) / PERCENTAGE_FACTOR
+        //     or x > (type(uint256).max - percentage / 2) / PERCENTAGE_FACTOR
         assembly {
             y := div(percentage, 2) // Temporary assignment to save gas.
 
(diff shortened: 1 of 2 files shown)


test/TestPercentageMath.sol
```diff
@@ -82,18 +82,21 @@ contract TestPercentageMath is Test {
...
-        vm.assume(x <= MAX_UINT256 / (PERCENTAGE_FACTOR + y));
+    function testPercentAdd(uint256 x, uint256 p) public {
+        vm.assume(p <= MAX_UINT256 - PERCENTAGE_FACTOR);
+        vm.assume(x <= MAX_UINT256 / (PERCENTAGE_FACTOR + p));
 
-        assertApproxEqAbs(PercentageMath.percentAdd(x, y), (x * (PERCENTAGE_FACTOR + y)) / PERCENTAGE_FACTOR, 1);
+        assertEq(
+            PercentageMath.percentAdd(x, p),
+            (x * (PERCENTAGE_FACTOR + p) + HALF_PERCENTAGE_FACTOR) / PERCENTAGE_FACTOR
+        );
```
MerlinEgalite, line 92: I would write those computations in a ref lib for maintainability

MerlinEgalite (changes requested):

src/math/PercentageMath.sol
```diff
@@ -41,36 +40,35 @@ library PercentageMath {
...
+        // percentage > PERCENTAGE_FACTOR
+        // or x * (PERCENTAGE_FACTOR - percentage) + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        // <=> percentage > PERCENTAGE_FACTOR
+        // or ((PERCENTAGE_FACTOR - percentage) > 0 and x > type(uint256).max - HALF_PERCENTAGE_FACTOR / (PERCENTAGE_FACTOR - percentage))
         assembly {
-            y := sub(PERCENTAGE_FACTOR, percentage) // Temporary assignement to save gas.
+            y := sub(PERCENTAGE_FACTOR, percentage) // Temporary assignment to save gas.
 
-            if mul(y, or(gt(percentage, PERCENTAGE_FACTOR), gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, y)))) {
+            if or(gt(percentage, PERCENTAGE_FACTOR), and(y, gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, y)))) {
```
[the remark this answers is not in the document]
MathisGD: `and(y, ...)` is always false if y == 0. This is what we want.
But you are right, there is an issue: y can be != 0, but `and(y, 1)` can be zero ! (for example y = 2)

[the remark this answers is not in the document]
MerlinEgalite: Perhaps we should add a comment then? Less clear to me and does not save gas either compare to the `mul`.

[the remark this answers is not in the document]
MathisGD: It saves some gas 
But we will come back to the mul, for the reason I mentioned yes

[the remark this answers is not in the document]
MathisGD: I found why the tests was passing: this branch was not tested !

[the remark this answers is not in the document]
MathisGD: fixed

test/TestPercentageMath.sol
MathisGD: I used mathRef

src/math/PercentageMath.sol
```diff
@@ -20,19 +20,18 @@ library PercentageMath {
...
+            y := add(PERCENTAGE_FACTOR, percentage) // Temporary assignment to save gas.
 
-            if mul(
-                y,
-                or(
-                    gt(percentage, sub(MAX_UINT256, PERCENTAGE_FACTOR)),
-                    gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, y))
-                )
+            if or(
+                gt(percentage, sub(MAX_UINT256, PERCENTAGE_FACTOR)),
```
Rubilmax, line 33: `sub(MAX_UINT256, PERCENTAGE_FACTOR)` could be extracted to a constant

src/math/PercentageMath.sol
```diff
@@ -45,15 +45,15 @@ library PercentageMath {
...
-        // Let percentage > 0
-        // Underflow if percentage > PERCENTAGE_FACTOR
-        // Overflow if x * (PERCENTAGE_FACTOR - percentage) + HALF_PERCENTAGE_FACTOR > type(uint256).max
-        // <=> x * (PERCENTAGE_FACTOR - percentage) > type(uint256).max - HALF_PERCENTAGE_FACTOR
-        // <=> x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR - percentage)
+        // Should revert if
+        // percentage > PERCENTAGE_FACTOR
+        // or x * (PERCENTAGE_FACTOR - percentage) + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        // <=> percentage > PERCENTAGE_FACTOR
+        // or ((PERCENTAGE_FACTOR - percentage) > 0 and x > type(uint256).max - HALF_PERCENTAGE_FACTOR / (PERCENTAGE_FACTOR - percentage))
```
Rubilmax, line 52:
```suggestion
        // or ((PERCENTAGE_FACTOR - percentage) > 0 and x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR - percentage))
```

Rubilmax: you dropped the parentheses?

test/TestPercentageMath.sol
```diff
@@ -159,15 +165,28 @@ contract TestPercentageMath is Test {
...
+    ) public {
+        vm.assume(percentage <= PERCENTAGE_FACTOR);
+        // prettier-ignore
+        vm.assume(
+            (percentage > 0 && y > MAX_UINT256_MINUS_HALF_PERCENTAGE / percentage) ||
+            (PERCENTAGE_FACTOR - percentage > 0 && x > (MAX_UINT256_MINUS_HALF_PERCENTAGE / (PERCENTAGE_FACTOR - percentage)))
         );
+
+        PercentageMath.weightedAvg(x, y, percentage);
     }
```
Rubilmax, line 190: this test should not pass right?

src/math/PercentageMath.sol
MathisGD: yes, but the compiler does a good job here and it is not saving any gas.
I find it more readable this way. And you ?

test/TestPercentageMath.sol
MathisGD: hmm, we assume x and such as it should fail
and we expect it to fail 
it seems ok

test/TestPercentageMath.sol
```diff
@@ -113,10 +110,7 @@ contract TestPercentageMath is Test {
         vm.assume(p <= PERCENTAGE_FACTOR);
         vm.assume(x <= MAX_UINT256 / (PERCENTAGE_FACTOR - p));
 
-        assertEq(
-            PercentageMath.percentSub(x, p),
-            (x * (PERCENTAGE_FACTOR - p) + HALF_PERCENTAGE_FACTOR) / PERCENTAGE_FACTOR
-        );
+        assertEq(PercentageMath.percentSub(x, p), mathRef.percentSub(x, p));
```
MerlinEgalite, line 113: small warning here but at the left we're calling directly the lib while at the right we're walling the contract.

src/math/PercentageMath.sol
```diff
@@ -45,15 +45,15 @@ library PercentageMath {
...
+        // Must revert if
         // percentage > PERCENTAGE_FACTOR
         // or x * (PERCENTAGE_FACTOR - percentage) + HALF_PERCENTAGE_FACTOR > type(uint256).max
         // <=> percentage > PERCENTAGE_FACTOR
         // or ((PERCENTAGE_FACTOR - percentage) > 0 and x > type(uint256).max - HALF_PERCENTAGE_FACTOR / (PERCENTAGE_FACTOR - percentage))
         assembly {
             y := sub(PERCENTAGE_FACTOR, percentage) // Temporary assignment to save gas.
 
-            if or(gt(percentage, PERCENTAGE_FACTOR), and(y, gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, y)))) {
+            if or(gt(percentage, PERCENTAGE_FACTOR), mul(y, gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, y)))) {
```
MerlinEgalite, line 56: thx

MerlinEgalite (approved):
A few comments but overall LGTM!

test/TestPercentageMath.sol
MathisGD: it is to avoid re-writing the ref two times. what do you prefer ?

src/math/PercentageMath.sol
MathisGD: fixed

src/math/PercentageMath.sol
Rubilmax: Oh, that's right you already told me this
Well then the only argument would be to stick to previous convention (we use this for `MAX_UINT256_MINUS_HALF_PERCENTAGE`), but honestly it's not important to me. Perhaps this one is more readable

Resolved!

test/TestPercentageMath.sol
Rubilmax: Naming `testFail` is a way to expect the test to revert??

src/math/PercentageMath.sol
```diff
@@ -100,45 +99,41 @@ library PercentageMath {
...
+        // percentage > PERCENTAGE_FACTOR
+        // or
+        // x1 = x * (PERCENTAGE_FACTOR - percentage) + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        // <=> (PERCENTAGE_FACTOR - percentage) > 0 and x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR - percentage)
+        // or
+        // x2 = y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
         // <=> y * percentage > type(uint256).max - HALF_PERCENTAGE_FACTOR
-        // <=> y > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
+        // <=> percentage > 0 and y > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
         //
```
QGarchery, line 121: It is really difficult to parse the order of the logical operators here. We can be more explicit and add a "Must also revert if" in between

src/math/PercentageMath.sol
```diff
@@ -100,45 +99,41 @@ library PercentageMath {
...
-        // Overflow if x1 + x2 > type(uint256).max
+        // Note: x1 / PERCENTAGE_FACTOR + x2 / PERCENTAGE_FACTOR < max(x,y) <= type(uint256).max
```
QGarchery, line 122: Let's write `PF` for `PERCENTAGE_FACTOR` and `HPF` for `HALF_PERCENTAGE_FACTOR`.

I don't think this note is true, for example this quantity is equal to the max when `x = y` and `percentage = PF`.
It can even be strictly greater than the max. For example with the values `x = y = 5000` and `percentage = 1`, we have `x1 / PF == x` and `x2 / PF == 1`

I still think there is no overflow when adding at the end (so this PR is safe), for the following reason. Let's assume that `percentage` is not equal to 0 and not equal to `PF`, otherwise it's obvious that there is no overflow because we would add 0. Then, because we checked overflow on `x1` and `x2`, we know that `x <= type(uint256).max - HPF` and same thing for `y`. On the other hand, we have a upper bound for `x1 / PF` given by `x * (1-p) + 1` where this last value is given in $\mathbb{R}$. This is obtained by changing `HPF` into `PF` in `x1`. Similarly, we have `x2 / PF` that is lower than `y * p +1`.

So `x1 / PF + x2 / PF <= x * (1-p) + 1 + y * p + 1 <= max(x, y) + 2 <= type(uint256).max - HPF + 2 <= type(uint256).max`

QGarchery (commented):
Nice one, there are some tricky things going on such as the use of bitwise `OR` when checking conditions (which works, contrary to the bitwise `AND`). Some things to clarify also

test/TestPercentageMath.sol
MerlinEgalite: as you want, but seems weird

test/TestPercentageMath.sol
MerlinEgalite: I faced some issues with foundry and the `vm.expectRevert` so perhaps @MathisGD can change it, just saying.

test/TestPercentageMath.sol
MathisGD: Ok, I will keep it like this to avoid code duplication

src/math/PercentageMath.sol
MathisGD: I think I do prefer `sub(MAX, div(PERCENTAGE, 2)`

test/TestPercentageMath.sol
MathisGD: Ok I can change it if you prefer. I don't mind

src/math/PercentageMath.sol
MerlinEgalite: don't really care about what we chose but definitely we should harmonize

src/math/PercentageMath.sol
MerlinEgalite: Perhaps we can keep the previous version, or tweak it

MathisGD: Indentation seems to be a good idea here

MerlinEgalite (approved):
LGTM

src/math/PercentageMath.sol
MathisGD: Let’s do this in a other pr

src/math/PercentageMath.sol
MathisGD: Ok, great that you dived on this, I didn't have in mind this edge case. 
But it makes me realize that this function might have an unwanted behavior : weighted average of a and b is not in [a,b]

src/math/PercentageMath.sol
```diff
@@ -100,45 +99,42 @@ library PercentageMath {
...
+        //     x2 = y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        //     <=> y * percentage > type(uint256).max - HALF_PERCENTAGE_FACTOR
+        //     <=> percentage > 0 and y > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
         //
-        // Overflow if x2 = y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
-        // <=> y * percentage > type(uint256).max - HALF_PERCENTAGE_FACTOR
-        // <=> y > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
-        //
-        // Overflow if x1 + x2 > type(uint256).max
+        // Note: x1 / PERCENTAGE_FACTOR + x2 / PERCENTAGE_FACTOR <= x * (1 - percentage) + 1 + y * percentage + 1
```
QGarchery, line 122:
```suggestion
        // Note: x1 / PERCENTAGE_FACTOR + x2 / PERCENTAGE_FACTOR <= x * (1 - p) + 1 + y * p + 1 where p is the fractional number percentage / PERCENTAGE_FACTOR
```
`p` is not the same as `percentage`

QGarchery (approved):
One thing to change then it's all good

QGarchery (approved):

PR #43, opened 2022-08-20
Fix weighted average double round up

MathisGD:
Fixes [this](https://github.com/morpho-dao/morpho-utils/pull/41#discussion_r946620510) issue found by @QGarchery.
Unexpected, but it is saving gas !
Notice that the reverts for overflows are much more conservative with this solution, but I don't think that this is much of an issue.

### DIFF
--- a/src/math/PercentageMath.sol
+++ b/src/math/PercentageMath.sol
@@ -112,12 +112,9 @@ library PercentageMath {
         // Must revert if
         //     percentage > PERCENTAGE_FACTOR
         // or if
-        //     x * (PERCENTAGE_FACTOR - percentage) > type(uint256).max
-        //     <=> PERCENTAGE_FACTOR - percentage > 0 and x > type(uint256).max / (PERCENTAGE_FACTOR - percentage)
-        // or if
         //     y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
         //     <=> percentage > 0 and y > (type(uint256).max - percentage) / percentage
-        // or if (assuming that the 3 previous conditions are false)
+        // or if
         //     x * (PERCENTAGE_FACTOR - percentage) + y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
         //     <=> x > type(uint256).max - y * percentage - HALF_PERCENTAGE_FACTOR / (PERCENTAGE_FACTOR - percentage)
         assembly {
@@ -125,11 +122,8 @@ library PercentageMath {
             if or(
                 gt(percentage, PERCENTAGE_FACTOR),
                 or(
-                    or(
-                        mul(percentage, gt(y, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, percentage))),
-                        mul(z, gt(x, div(MAX_UINT256, z)))
-                    ),
-                    gt(mul(x, z), sub(sub(MAX_UINT256, mul(y, percentage)), HALF_PERCENTAGE_FACTOR))
+                    mul(percentage, gt(y, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, percentage))),
+                    mul(z, gt(x, div(sub(sub(MAX_UINT256, mul(y, percentage)), HALF_PERCENTAGE_FACTOR), z)))
                 )
             ) {
                 revert(0, 0)
--- a/src/math/PercentageMath.sol
+++ b/src/math/PercentageMath.sol
@@ -112,29 +112,29 @@ library PercentageMath {
         // Must revert if
         //     percentage > PERCENTAGE_FACTOR
         // or if
-        //     x1 = x * (PERCENTAGE_FACTOR - percentage) + HALF_PERCENTAGE_FACTOR > type(uint256).max
-        //     <=> (PERCENTAGE_FACTOR - percentage) > 0 and x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR - percentage)
+        //     x * (PERCENTAGE_FACTOR - percentage) > type(uint256).max
+        //     <=> PERCENTAGE_FACTOR - percentage > 0 and x > type(uint256).max / (PERCENTAGE_FACTOR - percentage)
         // or if
-        //     x2 = y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
-        //     <=> y * percentage > type(uint256).max - HALF_PERCENTAGE_FACTOR
-        //     <=> percentage > 0 and y > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
-        //
-        // Note: x1 / PERCENTAGE_FACTOR + x2 / PERCENTAGE_FACTOR <= x * (1 - p) + 1 + y * p + 1 where p is the fractional number percentage / PERCENTAGE_FACTOR
-        // <= max(x, y) + 2 <= type(uint256).max - HALF_PERCENTAGE_FACTOR + 2 <= type(uint256).max
+        //     y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        //     <=> percentage > 0 and y > (type(uint256).max - percentage) / percentage
+        // or if (assuming that the 3 previous conditions are false)
+        //     x * (PERCENTAGE_FACTOR - percentage) + y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        //     <=> x > type(uint256).max - y * percentage - HALF_PERCENTAGE_FACTOR / (PERCENTAGE_FACTOR - percentage)
         assembly {
             z := sub(PERCENTAGE_FACTOR, percentage) // Temporary assignment to save gas.
-
             if or(
-                or(gt(percentage, PERCENTAGE_FACTOR), mul(z, gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, z)))),
-                mul(percentage, gt(y, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, percentage)))
+                gt(percentage, PERCENTAGE_FACTOR),
+                or(
+                    or(
+                        mul(percentage, gt(y, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, percentage))),
+                        mul(z, gt(x, div(MAX_UINT256, z)))
+                    ),
+                    gt(mul(x, z), sub(sub(MAX_UINT256, mul(y, percentage)), HALF_PERCENTAGE_FACTOR))
+                )
             ) {
                 revert(0, 0)
             }
-
-            z := add(
-                div(add(mul(x, z), HALF_PERCENTAGE_FACTOR), PERCENTAGE_FACTOR),
-                div(add(mul(y, percentage), HALF_PERCENTAGE_FACTOR), PERCENTAGE_FACTOR)
-            )
+            z := div(add(add(mul(x, z), mul(y, percentage)), HALF_PERCENTAGE_FACTOR), PERCENTAGE_FACTOR)
         }
     }
 }
--- a/src/math/PercentageMath.sol
+++ b/src/math/PercentageMath.sol
@@ -116,14 +116,14 @@ library PercentageMath {
         //     <=> percentage > 0 and y > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
         // or if
         //     x * (PERCENTAGE_FACTOR - percentage) + y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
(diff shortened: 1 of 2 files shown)


test/TestPercentageMath.sol
```diff
@@ -201,6 +200,17 @@ contract TestPercentageMath is Test {
         PercentageMath.weightedAvg(x, y, percentage);
     }
 
+    function testWeightedAvgOutOfBounds() public {
```
MerlinEgalite, line 203: cannot be fuzzed?

test/TestPercentageMath.sol
```diff
@@ -164,12 +167,9 @@ contract TestPercentageMath is Test {
...
     ) public {
-        vm.assume(percentage <= PERCENTAGE_FACTOR);
-        // prettier-ignore
-        vm.assume(
-            (percentage == 0 || y <= MAX_UINT256_MINUS_HALF_PERCENTAGE / percentage) &&
-            (PERCENTAGE_FACTOR - percentage == 0 || x <= MAX_UINT256_MINUS_HALF_PERCENTAGE / (PERCENTAGE_FACTOR - percentage))
-        );
+        vm.assume(percentage <= 1e4);
+        vm.assume(percentage == 0 || y <= (type(uint256).max - 0.5e4) / percentage);
+        vm.assume(1e4 - percentage == 0 || x <= (type(uint256).max - y * percentage - 0.5e4) / (1e4 - percentage));
```
MerlinEgalite, line 172: at the opposite of what is done elsewhere

MerlinEgalite (approved):
Some comments and questions but overall LGTM

test/TestPercentageMath.sol
MathisGD: you mean the literal values instead of using the constants ?

test/TestPercentageMath.sol
MathisGD: we could yes

MerlinEgalite (approved):
LGTM

src/math/PercentageMath.sol
```diff
@@ -112,29 +112,23 @@ library PercentageMath {
...
+        //     <=> percentage > 0 and y > (type(uint256).max - percentage) / percentage
         // or if
-        //     x2 = y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
-        //     <=> y * percentage > type(uint256).max - HALF_PERCENTAGE_FACTOR
-        //     <=> percentage > 0 and y > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
-        //
-        // Note: x1 / PERCENTAGE_FACTOR + x2 / PERCENTAGE_FACTOR <= x * (1 - p) + 1 + y * p + 1 where p is the fractional number percentage / PERCENTAGE_FACTOR
-        // <= max(x, y) + 2 <= type(uint256).max - HALF_PERCENTAGE_FACTOR + 2 <= type(uint256).max
+        //     x * (PERCENTAGE_FACTOR - percentage) + y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        //     <=> x > type(uint256).max - y * percentage - HALF_PERCENTAGE_FACTOR / (PERCENTAGE_FACTOR - percentage)
```
QGarchery, line 119:
```suggestion
        //     <=> PERCENTAGE_FACTOR - percentage > 0 and  x > (type(uint256).max - y * percentage - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR - percentage)
```

src/math/PercentageMath.sol
```diff
@@ -112,29 +112,23 @@ library PercentageMath {
...
         assembly {
             z := sub(PERCENTAGE_FACTOR, percentage) // Temporary assignment to save gas.
-
             if or(
-                or(gt(percentage, PERCENTAGE_FACTOR), mul(z, gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, z)))),
-                mul(percentage, gt(y, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, percentage)))
+                gt(percentage, PERCENTAGE_FACTOR),
+                or(
+                    mul(percentage, gt(y, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, percentage))),
+                    mul(z, gt(x, div(sub(sub(MAX_UINT256, mul(y, percentage)), HALF_PERCENTAGE_FACTOR), z)))
```
QGarchery, line 126: We can swap the order of sub to have a substraction of a constant by a constant. It should save some gas

src/math/PercentageMath.sol
```diff
@@ -112,29 +112,23 @@ library PercentageMath {
         // Must revert if
         //     percentage > PERCENTAGE_FACTOR
         // or if
-        //     x1 = x * (PERCENTAGE_FACTOR - percentage) + HALF_PERCENTAGE_FACTOR > type(uint256).max
-        //     <=> (PERCENTAGE_FACTOR - percentage) > 0 and x > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR - percentage)
+        //     y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        //     <=> percentage > 0 and y > (type(uint256).max - percentage) / percentage
```
QGarchery, line 116:
```suggestion
        //     <=> percentage > 0 and y > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
```

test/TestPercentageMath.sol
```diff
@@ -165,10 +168,10 @@ contract TestPercentageMath is Test {
         uint16 percentage
     ) public {
         vm.assume(percentage <= PERCENTAGE_FACTOR);
-        // prettier-ignore
+        vm.assume(percentage == 0 || y <= (MAX_UINT256 - HALF_PERCENTAGE_FACTOR) / percentage);
         vm.assume(
-            (percentage == 0 || y <= MAX_UINT256_MINUS_HALF_PERCENTAGE / percentage) &&
-            (PERCENTAGE_FACTOR - percentage == 0 || x <= MAX_UINT256_MINUS_HALF_PERCENTAGE / (PERCENTAGE_FACTOR - percentage))
+            PERCENTAGE_FACTOR - percentage == 0 ||
+                x <= (MAX_UINT256 - y * percentage - HALF_PERCENTAGE_FACTOR) / (PERCENTAGE_FACTOR - percentage)
```
QGarchery, line 174: The following is a general comment, and something I've been thinking about for the past weeks. So don't take it too litterally.

Are the test conditions really useful ? I mean if we rewrite them everytime we change the function, can we really say that it helped us maintain a certain behavior ? It is making development more difficult though, because we have to rewrite this test. I'm leaning more and more towards test that are more focused and durable, leaving behind completeness, as tests cannot be complete anyway. 

The general reasoning is that if we can write code more easily, then we can also fix the code more easily. Imagine a scenario for the same PR, but where you had to rewrite 100 tests like these (it would probably take more than a week of work in total). In this case, would you think it would be worth it to change this function at all ?

For this specific test, maybe we can restrict it to arbitrary large values like $10^{36}$

test/TestPercentageMath.sol
```diff
@@ -201,6 +204,17 @@ contract TestPercentageMath is Test {
...
+    function testWeightedAvgOutOfBounds() public {
+        uint256 x = 5000;
+        uint256 y = 5000;
+        uint256 percentage = 1;
+
+        uint256 weightedAvg = PercentageMath.weightedAvg(x, y, percentage);
+
+        assertLe(x, weightedAvg);
+        assertLe(weightedAvg, y);
+    }
```
QGarchery, line 216: As an example, the idea of this test sounds really good on the other hand. Once we fixed the behavior in this PR we want to maintain it forever (or at least it should cost a bit to work around it)

QGarchery (commented):
Nice ! The first condition allows you to check the second which is really cool. I think it does not even break symmetry: a call to `weightedAvg(a, b p)` reverts iff `weightedAvg(b, a, PERCENTAGE_FACTOR - p)` reverts. Not sure how to check that in solidity but it can be done in certora quite easily

test/TestPercentageMath.sol
MathisGD: I asked me the same question.. this make few sense, because as you said I'm just re-writing the exact same conditions than in the code itself.

MathisGD: I'm not going to change everything in this PR, but let's open an issue for this, and fix it later on.

test/TestPercentageMath.sol
MathisGD: yes I agree

test/TestPercentageMath.sol
Rubilmax: Fair point ; I think it's a consequence of the fact that this test is a unit test whereas integration tests (those you describe) are much more generic
I agree it's taking more time to change the implementation of a function tested with unit tests (because of the issue you raised) ; though I think you're forgetting that these very tests are useful during development because they can (not always) shed light on changes and induced edge cases and forces the developer to double-check with unit testing
Most of the time, it's just double-checking
Sometimes (at least for me), it showed me some edge-cases

src/math/PercentageMath.sol
MathisGD: oh yes great idea

test/TestPercentageMath.sol
Rubilmax: These unit tests are not testing that the overall function works as expected, rather than each branch of the code are working as expected
I agree it's lacking an integration test which tests a subset of values, just as you described

MathisGD: >These unit tests are not testing that the overall function works as expected, rather than each branch of the code are working as expected

yes, but we mostly copy the conditions written in the code to separate these branches, so it is easy to do something wrong

QGarchery: Good points, let's discuss on the dedicated issue

src/math/PercentageMath.sol
```diff
@@ -112,29 +112,23 @@ library PercentageMath {
...
-        //     x2 = y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
-        //     <=> y * percentage > type(uint256).max - HALF_PERCENTAGE_FACTOR
+        //     y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
         //     <=> percentage > 0 and y > (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
-        //
-        // Note: x1 / PERCENTAGE_FACTOR + x2 / PERCENTAGE_FACTOR <= x * (1 - p) + 1 + y * p + 1 where p is the fractional number percentage / PERCENTAGE_FACTOR
-        // <= max(x, y) + 2 <= type(uint256).max - HALF_PERCENTAGE_FACTOR + 2 <= type(uint256).max
+        // or if
+        //     x * (PERCENTAGE_FACTOR - percentage) + y * percentage + HALF_PERCENTAGE_FACTOR > type(uint256).max
+        //     <=> (PERCENTAGE_FACTOR - percentage) and (x > type(uint256).max - HALF_PERCENTAGE_FACTOR - y * percentage ) / (PERCENTAGE_FACTOR - percentage)
```
QGarchery, line 119:
```suggestion
        //     <=> (PERCENTAGE_FACTOR - percentage) > 0 and x > (type(uint256).max - HALF_PERCENTAGE_FACTOR - y * percentage) / (PERCENTAGE_FACTOR - percentage)
```

src/math/PercentageMath.sol
```diff
@@ -112,29 +112,23 @@ library PercentageMath {
...
         assembly {
             z := sub(PERCENTAGE_FACTOR, percentage) // Temporary assignment to save gas.
-
             if or(
-                or(gt(percentage, PERCENTAGE_FACTOR), mul(z, gt(x, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, z)))),
-                mul(percentage, gt(y, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, percentage)))
+                gt(percentage, PERCENTAGE_FACTOR),
+                or(
+                    mul(percentage, gt(y, div(MAX_UINT256_MINUS_HALF_PERCENTAGE, percentage))),
+                    mul(z, gt(x, div(sub(sub(MAX_UINT256, HALF_PERCENTAGE_FACTOR), mul(y, percentage)), z)))
```
QGarchery, line 126: Maybe chose the same thing for the 2 lines, either `MAX_UINT256_MINUS_HALF_PERCENTAGE` or `sub(MAX_UINT256, HALF_PERCENTAGE_FACTOR)`

Rubilmax (approved):

QGarchery (approved):
LGTM ! I think it is correct now, meaning that it respects its specification given by "Executes a weighted average (x * (1 - p) + y * p), rounded up". As a consequence, it should be bound up by `max(x, y)` and bounded down by `min(x, y)`

PR #45, opened 2022-08-23
Optimize min and max

Vectorized:
Ain't much, but it's honest work. Saves around 8 gas.

Somehow `.gas-snapshot` is under `.gitignore`. Is this intended?

### DIFF
--- a/src/math/Math.sol
+++ b/src/math/Math.sol
@@ -10,7 +10,7 @@ library Math {
 
     function max(uint256 x, uint256 y) internal pure returns (uint256 z) {
         assembly {
-            z := xor(x, mul(xor(x, y), lt(x, y)))
+            z := xor(x, mul(xor(x, y), gt(y, x)))
         }
     }
 
--- a/src/math/Math.sol
+++ b/src/math/Math.sol
@@ -4,13 +4,13 @@ pragma solidity ^0.8.0;
 library Math {
     function min(uint256 x, uint256 y) internal pure returns (uint256 z) {
         assembly {
-            z := add(mul(x, lt(x, y)), mul(y, iszero(lt(x, y))))
+            z := xor(x, mul(xor(x, y), lt(y, x)))
         }
     }
 
     function max(uint256 x, uint256 y) internal pure returns (uint256 z) {
         assembly {
-            z := add(mul(x, gt(x, y)), mul(y, iszero(gt(x, y))))
+            z := xor(x, mul(xor(x, y), lt(x, y)))
         }
     }
 


QGarchery (approved):
Really cool, thanks for your contribution !

I think it saves 8 gas rather than 18, essentially there is a `MUL` and a `ISZERO` less in your version. Actually the stack management is also a bit easier and saves one `SWAP`, but those 3 gas are compensated by the fact that the compiler is pretty smart and reuses the `lt(x, y)` computation.

This makes me think that we could maybe also do, for the `min` for example, something like:
`add(x, mul(sub(y, x), lt(y, x)))`
But your version is better because it never relies on overflow/underflow.

Vectorized:
Oh, I simply compared the gas values of the tests before and after.

`xor`, being commutative may help ease stack juggling in certain cases when the function is inlined.

MerlinEgalite:
> Ain't much, but it's honest work. Probably saves around 18 gas?
> 
> Somehow `.gas-snapshot` is under `.gitignore`. Is this intended?

I will re-add it, thx to have reminded me of it!

Rubilmax:
> Normally it should create logs thx to a bot we set up that's why we removed the snapshot. Seems not to work on this PR, not sure why though. @Rubilmax do you have an idea?

Definitely need to dig this, I'll take the time

MerlinEgalite (approved):
Nice ⚡️

MathisGD (commented):
Great !

src/math/Math.sol
```diff
@@ -4,13 +4,13 @@ pragma solidity ^0.8.0;
...
         assembly {
-            z := add(mul(x, lt(x, y)), mul(y, iszero(lt(x, y))))
+            z := xor(x, mul(xor(x, y), lt(y, x)))
         }
     }
 
     function max(uint256 x, uint256 y) internal pure returns (uint256 z) {
         assembly {
-            z := add(mul(x, gt(x, y)), mul(y, iszero(gt(x, y))))
+            z := xor(x, mul(xor(x, y), lt(x, y)))
```
MathisGD, line 13: I find it clearer, and more symmetrical with `min`
```suggestion
            z := xor(x, mul(xor(x, y), gt(y, x)))
```

MathisGD (approved):

QGarchery:
> Oh, I simply compared the gas values of the tests before and after.
>
> `xor`, being commutative may help ease stack juggling in certain cases when the function is inlined.

Yes, that is what I did too first:
before changes
before
after changes
after

Rubilmax (approved):

PR #46, opened 2022-08-28
Remove constants

### LINKED ISSUE #42: Harmonize `div` and `mul` between constants
`div` and `mul` between constants are optimized by the compiler and does not cost more gas than using constants. Perhaps we should come back to `div(MAX_UINT256, HALF_PERCENTAGE_FACTOR)`?

MerlinEgalite:
Fixes #42

### DIFF
--- a/src/math/CompoundMath.sol
+++ b/src/math/CompoundMath.sol
@@ -8,6 +8,7 @@ pragma solidity ^0.8.0;
 library CompoundMath {
     /// CONSTANTS ///
 
+    uint256 internal constant MAX = 2**256 - 1;
     uint256 public constant SCALE = 1e36;
     uint256 public constant WAD = 1e18;
 
@@ -17,7 +18,7 @@ library CompoundMath {
         assembly {
             // Revert if x * y > type(uint256).max
             // <=> y > 0 and x > type(uint256).max / y
-            if mul(y, gt(x, div(not(0), y))) {
+            if mul(y, gt(x, div(MAX, y))) {
                 revert(0, 0)
             }
 
@@ -29,7 +30,7 @@ library CompoundMath {
         assembly {
             // Revert if x * SCALE > type(uint256).max or y == 0
             // <=> x > type(uint256).max / SCALE or y == 0
-            if iszero(mul(y, iszero(gt(x, div(not(0), SCALE))))) {
+            if iszero(mul(y, iszero(gt(x, div(MAX, SCALE))))) {
                 revert(0, 0)
             }
 
--- a/src/math/CompoundMath.sol
+++ b/src/math/CompoundMath.sol
@@ -8,7 +8,6 @@ pragma solidity ^0.8.0;
 library CompoundMath {
     /// CONSTANTS ///
 
-    uint256 internal constant MAX_UINT256 = 2**256 - 1;
     uint256 public constant SCALE = 1e36;
     uint256 public constant WAD = 1e18;
 
@@ -18,7 +17,7 @@ library CompoundMath {
         assembly {
             // Revert if x * y > type(uint256).max
             // <=> y > 0 and x > type(uint256).max / y
-            if mul(y, gt(x, div(MAX_UINT256, y))) {
+            if mul(y, gt(x, div(not(0), y))) {
                 revert(0, 0)
             }
 
@@ -30,7 +29,7 @@ library CompoundMath {
         assembly {
             // Revert if x * SCALE > type(uint256).max or y == 0
             // <=> x > type(uint256).max / SCALE or y == 0
-            if iszero(mul(y, iszero(gt(x, div(MAX_UINT256, SCALE))))) {
+            if iszero(mul(y, iszero(gt(x, div(not(0), SCALE))))) {
                 revert(0, 0)
             }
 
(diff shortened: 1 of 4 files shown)


src/math/PercentageMath.sol
```diff
@@ -10,8 +10,6 @@ library PercentageMath {
 
     uint256 internal constant PERCENTAGE_FACTOR = 1e4; // 100.00%
```
MathisGD, line 11: IMO, this name is not perfect.. But I think that we want to have the same interface than Aave here right ?

src/math/PercentageMath.sol
```diff
@@ -10,8 +10,6 @@ library PercentageMath {
...
     uint256 internal constant HALF_PERCENTAGE_FACTOR = 0.5e4; // 50.00%
```
MathisGD, line 12: What is the reason behind not using `div(PERCENTAGE_FACTOR, 2)`

src/math/WadRayMath.sol
```diff
@@ -14,9 +14,6 @@ library WadRayMath {
     uint256 internal constant HALF_RAY = 0.5e27;
```
MathisGD, line 14: same question here

src/math/PercentageMath.sol
MerlinEgalite: yes

MathisGD: Ok

src/math/PercentageMath.sol
MerlinEgalite: because Aave is already doing it, so I  guess we can keep it

src/math/WadRayMath.sol
MerlinEgalite: same

src/math/CompoundMath.sol
```diff
@@ -18,7 +17,7 @@ library CompoundMath {
         assembly {
             // Revert if x * y > type(uint256).max
             // <=> y > 0 and x > type(uint256).max / y
-            if mul(y, gt(x, div(MAX_UINT256, y))) {
```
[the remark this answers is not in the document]
MerlinEgalite: yup we can change it

src/math/PercentageMath.sol
MerlinEgalite: I mean we should at least expose the same function and getters

MathisGD: ok ok

Rubilmax (commented):
The thing is, these libraries could be used with older versions of Solidity, if ever needed
And nothing ensures previous versions of the Solidity compiler did such constants optimizations (`sub(constant, constant) => constant`)

Also, did we check if one benefits from such optimizations without the optimization setting of the compiler?

TLDR; we should lower the solidity pragma of these libraries and don't rely on the compiler to optimize our code (just for the sake of beautiful code with less explicit constants) => we should explicitely define all such "optimization" implicit constants

MerlinEgalite:
> The thing is, these libraries could be used with older versions of Solidity, if ever needed And nothing ensures previous versions of the Solidity compiler did such constants optimizations (`sub(constant, constant) => constant`)
> 
> Also, did we check if one benefits from such optimizations without the optimization setting of the compiler?
> 
> TLDR; we could lower the solidity pragma of these libraries and we shouldn't rely on the compiler to optimize our code (just for the sake of beautiful code with less explicit constants) => we should explicitely define all such "optimization" implicit constants

This is a good point. We might not merge this PR then.

MathisGD:
I don't think that relying on newest versions of solidity for some optimizations is not an issue. It is quite common to simplify you code when the solidity compiler changes its behavior (see Solmate, relying on solidity integrated overflow management). Moreover, the number of people developing with really old versions of solidity is very small.

MerlinEgalite:
Yes I didn't the see the bots. Seems not to be an improvement, we must check in local too.

Good points too Mathis. Unless this discussion is resolved, I'll keep this PR pending.

MathisGD:
Ok so the gas cost is worse than before if the number of optimizations of the compiler is low...

MathisGD:
After thinking about it and discussing with part of the team, I think that we should optimize our code for a high number of optimization. So I would change the default number of optimizations of this repo, and merge it.

Rubilmax (dismissed):
What's the point in making these changes?

We're making optimization implicit, for the sake of the beauty of the code. I disagree with this move

MathisGD:
Readability (which helps maintaining a safe code)

MathisGD:
What should we do with this one ?

src/math/CompoundMath.sol
```diff
@@ -8,7 +8,7 @@ pragma solidity ^0.8.0;
 library CompoundMath {
     /// CONSTANTS ///
 
-    uint256 internal constant MAX_UINT256 = 2**256 - 1;
+    uint256 internal constant MAX = 2**256 - 1;
     uint256 public constant SCALE = 1e36;
```
pakim249CAL, line 12: Unrelated, but maybe we should be declaring these as internal too. Declaring a library variable as public would technically make this an external library.

pakim249CAL (approved):

Rubilmax:
I still disagree with this change, as I don't think removing explicit constants increases the readability of the code (it even decreases it IMO) lol

MathisGD:
Are you sure that you prefer
```solidity
MAX_MINUS_HALF_RAY
```

over

```solidity
sub(MAX, div(RAY, 2))
```

It is not obvious to me

PR #47, opened 2022-08-29
⬆️ Update foundry-gas-diff

Rubilmax:


.github/workflows/foundry-gas-diff.yml
```diff
@@ -28,13 +28,16 @@ jobs:
           version: nightly
 
       - name: Run tests
-        run: forge test --gas-report > gasreport.ansi
+        run: forge test --gas-report | tee gasreport.ansi
+        env:
+          # make fuzzing semi-deterministic to avoid noisy gas cost estimation
+          # due to non-deterministic fuzzing, but keep it not always deterministic
+          FOUNDRY_FUZZ_SEED: 0x${{ github.event.pull_request.base.sha || github.sha }}
```
MathisGD, line 35: Idk if this is an issue in this repo 🤔

Rubilmax: It's not an issue currenlty but still, we expect the CI to be reproducible

MathisGD: OK but imagine this seed does not test a edge case but an other do, we could miss it

MerlinEgalite: Yes agree with the last argument

MerlinEgalite: Although the CI seems to fail [here](https://github.com/morpho-dao/morpho-utils/pull/46).

Rubilmax: We'd miss it on this PR, but not on another or when the CI runs on dev/main (hence the goal of keeping it non-deterministic)...

MerlinEgalite (approved):
LGTM

.github/workflows/foundry-gas-diff.yml
MathisGD: so we keep the deterministic CI ?

PR #54, opened 2022-11-04
Add memory-safe NatSpec tag to assembly blocks

Rubilmax:
This PR requires the reviewer to be aware of the following:
- https://docs.soliditylang.org/en/v0.8.17/assembly.html#memory-safety

### In particular, please note the Solidity documentation states:

> Inline assembly that neither involves any operations that access memory nor assigns to any Solidity variables in memory is automatically considered memory-safe and does not need to be annotated.

However, it seems to not be the case when I compile the following with Solidity 0.8.16 without the optimizer and with IR:

```solidity
contract Test {
  using CompountMath for uint256;

  function test() {
    uint256 a = uint256(1).mul(1 ether);
  }
}
```

Because I get the error:

```bash
Variable _2 is 1 too deep in the stack [ _2 expr_mpos_3 _6 _18 _1 _16 expr_component_mpos_1 _10 RET var_ins_mpos _4 _7 _3 _14 var_i expr_1 _29 ]
No memoryguard was present. Consider using memory-safe assembly only and annotating it via 'assembly ("memory-safe") { ... }'.
```

> In case you are developing a library that is meant to be compatible across multiple versions of Solidity, you can use a special comment to annotate an assembly block as memory-safe

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -1,7 +1,5 @@
 [profile.default]
-src = 'src'
-out = 'out'
-libs = ['lib']
+via-ir = true
 fuzz_runs = 1024
 gas_reports = ["*"]
 
--- a/src/math/CompoundMath.sol
+++ b/src/math/CompoundMath.sol
@@ -15,6 +15,7 @@ library CompoundMath {
     /// INTERNAL ///
 
     function mul(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        /// @solidity memory-safe-assembly
         assembly {
             // Revert if x * y > type(uint256).max
             // <=> y > 0 and x > type(uint256).max / y
@@ -27,6 +28,7 @@ library CompoundMath {
     }
 
     function div(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        /// @solidity memory-safe-assembly
         assembly {
             // Revert if x * SCALE > type(uint256).max or y == 0
             // <=> x > type(uint256).max / SCALE or y == 0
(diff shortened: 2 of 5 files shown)


MathisGD:
The gas cost isn't supposed to change right ?

MathisGD:
I don't have any issues on my side... even with 0.8.16. Moreover I don't find any mention of this in the Solidity 0.8.17. If there is no issue, I would prefer not adding them, as Solidity seems to do a great job at auto-flagging them.

Let's discuss this tomorrow.

MerlinEgalite:
> This PR requires the reviewer to be aware of the following:
> 
> * https://docs.soliditylang.org/en/v0.8.17/assembly.html#memory-safety
> 
> ### In particular, please note the Solidity documentation states:
> > Inline assembly that neither involves any operations that access memory nor assigns to any Solidity variables in memory is automatically considered memory-safe and does not need to be annotated.
> 
> However, it seems to not be the case when I compile the following with Solidity 0.8.16:
> 
> ```solidity
> contract Test {
>   using CompountMath for uint256;
> 
>   function test() {
>     uint256 a = uint256(1).mul(1 ether);
>   }
> }
> ```
> 
> Because I get the error:
> 
> ```shell
> Variable _2 is 1 too deep in the stack [ _2 expr_mpos_3 _6 _18 _1 _16 expr_component_mpos_1 _10 RET var_ins_mpos _4 _7 _3 _14 var_i expr_1 _29 ]
> No memoryguard was present. Consider using memory-safe assembly only and annotating it via 'assembly ("memory-safe") { ... }'.
> ```
> 
> > In case you are developing a library that is meant to be compatible across multiple versions of Solidity, you can use a special comment to annotate an assembly block as memory-safe

Did you check if a is stored in memory or not?

Rubilmax:
So the error vanished. I can't reproduce the bug, so I'm closing this PR

Issue #56, opened 2022-11-20
Formally prove `morpho-utils`


MathisGD:
- [x] math libraries
- [ ] delegate call libraries

MathisGD:
Closing, replaced by #91

PR #59, opened 2022-11-22
Formal verification (certora)

MathisGD:
Formal verification (#56) of:
- `Math.sol` ✅ 
- `CompoundMath.sol` ✅ 
- `WadRayMath.sol` ✅ 
- `PercentageMath.sol` ✅ 

To do: 
- [x] adapt README
- [x] add it to the CI


certora/specs/percentageMath.spec
```diff
@@ -0,0 +1,77 @@
...
+rule percentAddSafety(uint256 x, uint256 p) {
+    uint res = percentAdd(x, p);
+
+    assert res == (x * (10^4 + p) + 10^4 / 2) / 10^4;
+}
+
+rule percentAddLiveness(uint256 x, uint256 p) {
+    percentAdd@withrevert(x, p);
+
+    assert lastReverted <=> 10^4 + p >= 2^256 || x * (10^4 + p) + 10^4 / 2 >= 2^256;
```
QGarchery, line 20: Some comment on why those conditions make sense would be nice

certora/specs/math.spec
```diff
@@ -0,0 +1,79 @@
...
+methods {
+    min(uint256, uint256)           returns (uint256)       envfree
+    max(uint256, uint256)           returns (uint256)       envfree
+    zeroFloorSub(uint256, uint256)  returns (uint256)       envfree
+    divUp(uint256, uint256)         returns (uint256)       envfree
+}
+
+/// min ///
+
+rule minSafety(uint256 a, uint256 b) {
```
QGarchery, line 11: Rules are fixed when run with the flag ` --settings -useBitVectorTheory`

QGarchery (changes requested):
Awesome, just a few comments and fixes to make rules pass !

certora/run.sh
```diff
@@ -0,0 +1,29 @@
...
+    --send_only                                                     \
+    --msg "PercentageMath"                                          \
+    $@
+
+certoraRun                                                          \
+    certora/helpers/MockCompoundMath.sol                            \
+    --verify MockCompoundMath:certora/specs/compoundMath.spec       \
+    --send_only                                                     \
+    --msg "CompoundMath"                                            \
+    $@
```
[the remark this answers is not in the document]
MathisGD: It's useful for me because my solc setup is weird

certora/specs/compoundMath.spec
```diff
@@ -0,0 +1,34 @@
...
+    uint res = div(x, y);
+
+    assert res == x * WAD() / y; 
+}
+
+rule divLiveness(uint256 x, uint256 y) {
+    div@withrevert(x, y);
+
+    assert lastReverted <=> x * WAD() >= 2^256 || y == 0;
+}
```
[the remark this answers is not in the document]
MathisGD: I prefer to merge the fix

certora/specs/math.spec
MathisGD: oh amazing

certora/specs/wadRayMath.spec
```diff
@@ -0,0 +1,96 @@
...
+definition WAD()        returns uint = 10^18;
+definition RAY()        returns uint = 10^27;
+definition WADTORAY()   returns uint = 10^9;
+
+/// wadMul ///
+
+rule wadMulSafety(uint256 a, uint256 b) {
+    uint res = wadMul(a, b);
+
+    assert res == (a * b + (WAD() / 2)) / WAD();
```
QGarchery, line 19: It would be nice to also explicit the two cases. We could for example add the following assertion:
```solidity 
assert a * b % WAD() < WAD() / 2 => res == a * b / WAD();`

**Review by QGarchery — APPROVED** [review]
LGTM, some optional changes

File: certora/specs/math.spec [hunk 6]
```diff
@@ -0,0 +1,67 @@
...
+    uint res = min(a, b);
+
+    assert a <= b => res == a;
+    assert a > b => res == b;
+}
+
+rule minLiveness(uint256 a, uint256 b) {
+    min@withrevert(a, b);
+
+    assert lastReverted <=> false;
```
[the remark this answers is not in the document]
MathisGD: the current version is more homogenous with the rest

certora/specs/wadRayMath.spec
MathisGD: You mean here and everywhere else ?

Rubilmax:
certora is cool

certora/specs/wadRayMath.spec
MathisGD: I tried it, but I think that it is over complicating things

certora/helpers/MockWadRayMath.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import "../../src/math/WadRayMath.sol";
+
+contract MockWadRayMath {
```
MerlinEgalite, line 6: Those contracts already exist in the test suite I guess you can reuse them no?

certora/specs/compoundMath.spec
```diff
@@ -0,0 +1,34 @@
+methods {
+    mul(uint256, uint256)   returns (uint256)   envfree
+    div(uint256, uint256)   returns (uint256)   envfree
+}
+
+definition WAD()        returns uint = 10^18;
```
MerlinEgalite, line 6: wow what are those spaces? 👀

certora/specs/percentageMath.spec
```diff
@@ -0,0 +1,78 @@
+methods {
+    percentAdd(uint256, uint256)            returns (uint256)   envfree
+    percentSub(uint256, uint256)            returns (uint256)   envfree
+    percentMul(uint256, uint256)            returns (uint256)   envfree
+    percentDiv(uint256, uint256)            returns (uint256)   envfree
+    weightedAvg(uint256, uint256, uint256)  returns (uint256)   envfree
+}
+
```
MerlinEgalite, line 8:
```suggestion

definition MAX_BPT()             returns uint = 10^4;
definition HALF_MAX_BPT()        returns uint = 10^4;

```

MerlinEgalite (commented):
Some suggestions

certora/helpers/MockWadRayMath.sol
MathisGD: Yes I even think that we should remove them from the tests (they are only used for the gas report, which is not very useful)

MerlinEgalite: You mean replace them for gas snapshots?

MathisGD: yes ! It works much better

MerlinEgalite: I wouldn't say better because it's not testing the same thing at all. One allows you to check that your change is improving gas efficiency or not while the other is to compare gas from the reference which is a different goal IMO

MerlinEgalite: We can agree on getting rid of them, but we should be sure to understand what we're removing first

MathisGD: We could compare to reference with snapshot also

PR #70, opened 2022-12-13
Add safe cast (not optimized)

MerlinEgalite:
Actually, this PR is a bit disappointing since it does bring any gas improvements when it's not reverting 🥲

Perhaps someone else has good ideas

### DIFF
--- a/src/math/SafeCast.sol
+++ b/src/math/SafeCast.sol
@@ -7,11 +7,66 @@ pragma solidity ^0.8.0;
 /// @notice TODO
 library SafeCast {
 
+	uint256 internal constant MAX_UINT128 = 2**128 - 1; // Not possible to use type(uint128).max in yul
 	uint256 internal constant MAX_UINT96 = 2**96 - 1; // Not possible to use type(uint96).max in yul
+	uint256 internal constant MAX_UINT64 = 2**64 - 1; // Not possible to use type(uint64).max in yul
+	uint256 internal constant MAX_UINT32 = 2**32 - 1; // Not possible to use type(uint32).max in yul
+	uint256 internal constant MAX_UINT16 = 2**16 - 1; // Not possible to use type(uint16).max in yul
+	uint256 internal constant MAX_UINT8 = 2**8 - 1; // Not possible to use type(uint8).max in yul
+
+	function toUint128(uint256 value) internal pure returns(uint128 casted) {
+		assembly {
+			if iszero(lt(value, MAX_UINT128)) {
+				revert(0, 0)
+			}
+
+			casted := value
+		}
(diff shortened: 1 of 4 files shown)


Rubilmax:
I think the fact that there's no gas improvement comes down to the fact that Solidity type casting already translates to a sequence of opcodes that will be difficult to simplify further:
- when casting to a smaller type: it's only statically casting the variable (meaning there's no runtime behavior changes to the memory or the stack)
- when casting to a larger type: the bytes which were not used before are zero-ed out

Considering this, I'm against merging this PR

MerlinEgalite:
> I think the fact that there's no gas improvement comes down to the fact that Solidity type casting already translates to a sequence of opcodes that will be difficult to simplify further:
> 
> * when casting to a smaller type: it's only statically casting the variable (meaning there's no runtime behavior changes to the memory or the stack)
> * when casting to a larger type: the bytes which were not used before are zero-ed out
> 
> Considering this & the fact the these libraries are to be used inside Solidity contracts, I'm against merging this PR

It's not to be merged. This is just to showcase that it seems impossible to optimize.

Your answer is a response only for the casting case. The check could have been improved as we're doing the in other libraries. However, this is not the case as the compiler seems to already use the most optimized way to do it.

src/math/SafeCast.sol
```diff
@@ -0,0 +1,86 @@
+// SPDX-License-Identifier: GNU AGPLv3
+pragma solidity ^0.8.0;
+
+/// @title WadRayMath.
+/// @author Morpho Labs.
+/// @custom:contact security@morpho.xyz
+/// @notice TODO
+library SafeCast {
+
+	uint256 internal constant MAX_UINT128 = 2**128 - 1; // Not possible to use type(uint128).max in yul
```
MathisGD, line 10: Why is the tab so l a r g e ?

MerlinEgalite: the tab? Same as other repors no?

MathisGD: Ok I think that it was a bug on my side

src/math/SafeCast.sol
```diff
@@ -0,0 +1,86 @@
...
+				revert(0, 0)
+			}
+
+			casted := value
+		}
+	}
+
+	function toUint8(uint256 value) internal pure returns(uint8 casted) {
+		assembly {
+			if lt(MAX_UINT8, value) {
```
MathisGD, line 69: Issue here I think

MathisGD: The issue is still there I think

MathisGD:
Should we close this one ?

PR #72, opened 2022-12-15
Move Certora helpers to mocks

### LINKED ISSUE #71: Move mocks from `certora` and `test` folders to a common place
There are code duplications in those 2 folders I propose migrating them to a common folder to avoid code duplication.

Should we move them to the `test` folder?

Rubilmax:
Fixes #71 

- factorizing code shared between forge tests & certora tests
- only keeping forge test cases for gas tracking


.vscode/settings.json
```diff
@@ -1,11 +0,0 @@
-{
```
MerlinEgalite, line 1: what's the reason here?

test/TestCompoundMath.sol
```diff
@@ -1,93 +1,30 @@
...
-
-    function testGasMul() public view {
-        compoundMath.mul(2 * WAD, WAD);
-        compoundMathRef.mul(2 * WAD, WAD);
-    }
+        vm.assume(y > 0 && x < type(uint256).max / 1e36);
 
-    function testGasDiv() public view {
-        compoundMath.div(2 * WAD, WAD);
-        compoundMathRef.div(2 * WAD, WAD);
```
MerlinEgalite, line 91: why delete them?

test/TestCompoundMath.sol
```diff
@@ -1,93 +1,30 @@
...
 
     function testMul(uint256 x, uint256 y) public {
-        unchecked {
-            vm.assume(y == 0 || (x * y) / y == x);
-        }
-
-        assertEq(CompoundMath.mul(x, y), CompoundMathRef.mul(x, y));
-    }
-
-    function testMulOverflow(uint256 x, uint256 y) public {
```
MerlinEgalite, line 50: Is it delete because it's tested with certora?

MerlinEgalite (changes requested):

.vscode/settings.json
Rubilmax: These were messing with my setup (namely: prettier was not running)
I'm not a fan of sharing IDE settings, but if you require it then I'll try to fix this

QGarchery (approved):

pakim249CAL (approved):

.vscode/settings.json
MerlinEgalite: No it's ok

test/TestCompoundMath.sol
MerlinEgalite: But why `testMul` is kept then?

Btw one I thought about after is that external contributors may not be able to create certora tests and without testing example it might be difficult for them to add tests

Rubilmax: 1. For gas tracking: it allows to include the gas consumption into the gas report

2. Yes, and the argument also holds for the team: by introducing this change, we expect from contributors to develop & run tests using Certora only

I'm ok with that, as any open-source project relies on a specific tech stack which may not be easily mastered... But I don't see this as a viable argument against moving to Certora tests in this case, as we can safely state that Certora > foundry fuzzing

Also consider the origin of this PR: in order to add a feature to the repo, I had to write the exact same tests twice (once with Certora, once with forge)

MerlinEgalite: 1. For gas tracking ok but you're removing the gas comparison with the reference library right? I'm against it because one of the goals is to be sure that our lib is at least better than the reference one.

2. Yes but with Certora you need a key to run tests if I'm correct, while you don't need it for forge tests. So there are more constraints to writing certora tests than pure forge tests. Not a blocker to me, we can keep the current version but I think we must keep that in mind.

Rubilmax: > you're removing the gas comparison with the reference library right?

As you can see, I compare the result of our own library vs the result of the reference library. So the gas consumption of both libraries are tracked. Am I missing something?

> with Certora you need a key to run tests

This is something I wasn't aware of. We should discuss this then

MerlinEgalite: Ok I get it now. It's just that it seems to be a test and not a gas comparison check when reading it. So it's like if we were testing only this the correct behavior for this condition: `vm.assume(y == 0 || x < type(uint256).max / y);`. 

I would add Gas in the naming and remove: `/// TESTS ///` because it's misleading, no? (perhaps only on my side).

How did you run certora tests locally?

test/TestCompoundMath.sol
MathisGD: I think that they bring nothing

test/TestCompoundMath.sol
MathisGD: I think that we should keep both ! (We only remove testGas..)

MathisGD (changes requested):
We should keep all the tests (except gas ones) IMO

MathisGD:
Note that the reason why I did this is because it is classic to have to modify a little bit contracts in order to verify them formally. And it is usually done through a diff file in the `certoraˋ folder. 

But I don't see any issue with cleaning as you did it.

Rubilmax:
> We should keep all the tests (except gas ones) IMO

You're suggesting to keep the forge tests which are redundant with the Certora tests and remove the tests used to track the gas consumption. The result is:
1. A developer needs to write the exact same tests twice (once with Certora, once with forge)
2. We can't track the gas consumption of our library vs the reference library

This is neither what I suggest, nor what @MerlinEgalite suggests. Can you please provide arguments or insights to back your POV?

test/TestCompoundMath.sol
```diff
@@ -1,93 +1,49 @@
...
+        vm.assume(y == 0 || x <= type(uint256).max / y);
 
-        assertEq(CompoundMath.mul(x, y), CompoundMathRef.mul(x, y));
+        assertEq(mock.mul(x, y), ref.mul(x, y));
     }
 
     function testMulOverflow(uint256 x, uint256 y) public {
-        unchecked {
-            vm.assume(y > 0 && (x * y) / y != x);
-        }
```
MathisGD, line 53: This was also testing that the conditions were equivalent

test/TestWadRayMath.sol
```diff
@@ -68,132 +15,96 @@ contract TestWadRayMath is Test {
...
 
     function testRayToWad(uint256 x) public {
-        assertEq(WadRayMath.rayToWad(x), WadRayMathRef.rayToWad(x));
+        assertEq(mock.rayToWad(x), ref.rayToWad(x));
     }
 
     function testWadToRay(uint256 x) public {
-        unchecked {
-            vm.assume((x * WAD_RAY_RATIO) / WAD_RAY_RATIO == x);
-        }
```
MathisGD, line 154: same

MathisGD (approved):
A few remarks, but not blocking

test/TestCompoundMath.sol
Rubilmax: Indeed though I don't think we should include syntax tests inside our test suite?
I rather see this as a Solidity syntax test which should be included in a Solidity test suite

MerlinEgalite (approved):

PR #74, opened 2022-12-20
Optimized safeCast

MathisGD:
Optimize SafeCast, notably the non reverting case.


MerlinEgalite (commented):
Hmm did you try to use it in a real function with some logic after the safe cast? I'm not sure it would work

MathisGD:
Closing as it does not beat Solmate

PR #75, opened 2022-12-21
Logarithm

QGarchery:
Test and compare all possible implementations

### DIFF
--- a/src/math/Math.sol
+++ b/src/math/Math.sol
@@ -33,9 +33,11 @@ library Math {
         }
     }
 
+    /// @dev Returns the floor of log2(x) and returns 0 on input 0.
+    /// @dev This computation makes use of De Bruijn sequences, their usage for computing log2 is referenced here: http://supertech.csail.mit.edu/papers/debruijn.pdf
     function log2(uint256 x) internal pure returns (uint256 y) {
         assembly {
-            let arg := x
+            // Use the highest set bit of x to set each of its lower bits
             x := or(x, div(x, 0x02))
             x := or(x, div(x, 0x04))
             x := or(x, div(x, 0x10))
@@ -44,7 +46,11 @@ library Math {
             x := or(x, div(x, 0x100000000))
             x := or(x, div(x, 0x10000000000000000))
             x := or(x, div(x, 0x100000000000000000000000000000000))
+            // Take only the highest bit of x
+            x := xor(x, div(x, 0x02))
 
+            // Lookup table storing the 256 bytes obtained by "hashing" the powers of 2 against one De Bruijn sequence
+            // See the referenced article for a detailed explanation
             let m := mload(0x40)
             mstore(m, 0x0001020903110a19042112290b311a3905412245134d2a550c5d32651b6d3a75)
             mstore(add(m, 0x20), 0x06264262237d468514804e8d2b95569d0d495ea533a966b11c886eb93bc176c9)
@@ -56,9 +62,10 @@ library Math {
             mstore(add(m, 0xe0), 0xfd1e3e5a7a8aa2b670c4ced8bbe8f0f4fc3d79a1c3cde7effb78cce6facbf9f8)
             mstore(0x40, add(m, 0x100))
 
+            // This De Bruijn sequence begins with the byte 0, which is important to make shifting work like a rotation on the first byte
             let deBruijnSeq := 0x818283848586878898a8b8c8d8e8f929395969799a9b9d9e9faaeb6bedeeff
             let shift := 0x100000000000000000000000000000000000000000000000000000000000000
-            let key := div(mul(x, deBruijnSeq), shift)
+            let key := div(mul(x, deBruijnSeq), shift) // With the multiplication it works also for 0
             y := div(mload(add(m, key)), shift)
         }
     }
--- a/src/math/Math.sol
+++ b/src/math/Math.sol
@@ -63,7 +63,7 @@ library Math {
             mstore(0x40, add(m, 0x100))
 
             // This De Bruijn sequence begins with the byte 0, which is important to make shifting work like a rotation on the first byte
-            let deBruijnSeq := 0x818283848586878898a8b8c8d8e8f929395969799a9b9d9e9faaeb6bedeeff
+            let deBruijnSeq := 0x00818283848586878898a8b8c8d8e8f929395969799a9b9d9e9faaeb6bedeeff
             let key := shr(248, mul(x, deBruijnSeq)) // With the multiplication it works also for 0
             y := shr(248, mload(add(m, key))) // Look in the table and take the first byte
         }
--- a/src/math/Math.sol
+++ b/src/math/Math.sol
@@ -38,19 +38,19 @@ library Math {
     function log2(uint256 x) internal pure returns (uint256 y) {
         assembly {
             // Use the highest set bit of x to set each of its lower bits
-            x := or(x, shr(x, 1))
-            x := or(x, shr(x, 2))
-            x := or(x, shr(x, 4))
-            x := or(x, shr(x, 8))
-            x := or(x, shr(x, 16))
-            x := or(x, shr(x, 32))
-            x := or(x, shr(x, 64))
-            x := or(x, shr(x, 128))
+            x := or(x, shr(1, x))
+            x := or(x, shr(2, x))
+            x := or(x, shr(4, x))
+            x := or(x, shr(8, x))
+            x := or(x, shr(16, x))
+            x := or(x, shr(32, x))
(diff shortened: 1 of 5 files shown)


Morphojo (approved):
Ok for me

Morphojo (approved):
OK for me

MerlinEgalite (commented):
Awesome work! 🎉

As we discussed @QGarchery it would be nice to test it against a well-known python lib or something battle-tested.

test/helpers/DeBruijn.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: GNU AGPLv3
+pragma solidity ^0.8.0;
+
+library DeBruijn {
+    function isDeBruijnSequence(uint256 seq) public pure returns (bool) {
```
pakim249CAL, line 5: I would advise against using public library functions in general since the meaning of external is very different and often unexpected for libraries. This should only require internal.

pakim249CAL (approved):

test/helpers/DeBruijn.sol
```diff
@@ -0,0 +1,39 @@
...
+        bool[] memory seen = new bool[](256);
+
+        for (uint256 i; i < 256; i++) {
+            uint256 x = 2**i;
+            uint256 key;
+            // Inline assembly to avoid overflow
```
MathisGD, line 11: shifts to the right can overflow ?

test/helpers/DeBruijn.sol
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    function precomputeRows(uint256 seq) public pure returns (uint256[] memory) {
+        uint256[] memory rows = new uint256[](8);
+
+        for (uint256 i; i < 256; i++) {
+            uint256 x = 2**i;
+
+            uint256 key;
+            // Inline assembly to avoid overflow
```
MathisGD, line 28: Same question

MathisGD (approved):
It took me some time to understand ^^
Cool stuff ! Congrats @QGarchery

src/math/Math.sol
```diff
@@ -32,4 +32,40 @@ library Math {
...
+            let m := mload(0x40)
+            mstore(m, 0x0001020903110a19042112290b311a3905412245134d2a550c5d32651b6d3a75)
+            mstore(add(m, 0x20), 0x06264262237d468514804e8d2b95569d0d495ea533a966b11c886eb93bc176c9)
+            mstore(add(m, 0x40), 0x071727374353637324837e9b47af86c7155181ad4fd18ed32c9096db57d59ee3)
+            mstore(add(m, 0x60), 0x0e2e4a6a5f92a6be3498aae067ddb2eb1d5989b56fd7baf33ca0c2ee77e5caf7)
+            mstore(add(m, 0x80), 0xff0810182028303840444c545c646c7425617c847f8c949c48a4a8b087b8c0c8)
+            mstore(add(m, 0xa0), 0x16365272829aaec650acd0d28fdad4e22d6991bd97dfdcea58b4d6f29fede4f6)
+            mstore(add(m, 0xc0), 0xfe0f1f2f3f4b5b6b607b8b93a3a7b7bf357199c5abcfd9e168bcdee9b3f1ecf5)
+            mstore(add(m, 0xe0), 0xfd1e3e5a7a8aa2b670c4ced8bbe8f0f4fc3d79a1c3cde7effb78cce6facbf9f8)
+            mstore(0x40, add(m, 0x100))
```
Rubilmax, line 63:
```suggestion
            mstore(0x40, add(m, 0x100)) // Update the free memory pointer
```

test/TestMath.sol
```diff
@@ -53,4 +54,54 @@ contract TestMath is Test {
...
+
+        uint256[] memory rows = DeBruijn.precomputeRows(deBruijnSeq);
+        uint256 rowFromLookupTable;
+        for (uint256 i; i < 8; i++) {
+            assembly {
+                rowFromLookupTable := mload(add(m, mul(32, i)))
+            }
+            assertEq(rowFromLookupTable, rows[i]);
+        }
+    }
```
Rubilmax, line 106: If I'm not mistaken, we're running a test against a test helper library (we cannot call these unit testing because these are not integrated with the true production code and thus cannot defend the role of unit testing the library, because the library's code can be desynced with the test code)

I'm not a fan of these, but we've already discussed it. Could we just move them to another test file, so that we separate test file concerns (and keep `TestMath.sol` only testing `Math.sol`?)

MathisGD: It is still related to `Math.sol`, so it makes sense to me that it stays here.

But as you want, this is not important.

src/math/Math.sol
```diff
@@ -32,4 +32,40 @@ library Math {
...
+            // Let x be a power of 2, its hash is obtained by taking the first byte of x * deBruijnSeq
+            let m := mload(0x40)
+            mstore(m, 0x0001020903110a19042112290b311a3905412245134d2a550c5d32651b6d3a75)
+            mstore(add(m, 0x20), 0x06264262237d468514804e8d2b95569d0d495ea533a966b11c886eb93bc176c9)
+            mstore(add(m, 0x40), 0x071727374353637324837e9b47af86c7155181ad4fd18ed32c9096db57d59ee3)
+            mstore(add(m, 0x60), 0x0e2e4a6a5f92a6be3498aae067ddb2eb1d5989b56fd7baf33ca0c2ee77e5caf7)
+            mstore(add(m, 0x80), 0xff0810182028303840444c545c646c7425617c847f8c949c48a4a8b087b8c0c8)
+            mstore(add(m, 0xa0), 0x16365272829aaec650acd0d28fdad4e22d6991bd97dfdcea58b4d6f29fede4f6)
+            mstore(add(m, 0xc0), 0xfe0f1f2f3f4b5b6b607b8b93a3a7b7bf357199c5abcfd9e168bcdee9b3f1ecf5)
+            mstore(add(m, 0xe0), 0xfd1e3e5a7a8aa2b670c4ced8bbe8f0f4fc3d79a1c3cde7effb78cce6facbf9f8)
```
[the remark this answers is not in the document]
QGarchery: I love it, but can you also manage to make it easily reusable ? I tried passing it as calldata, but it was less efficient idk why.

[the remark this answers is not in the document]
QGarchery: Calldata attempt: https://github.com/morpho-dao/morpho-utils/pull/78

src/math/Math.sol
```diff
@@ -32,4 +32,40 @@ library Math {
...
+            x := or(x, shr(1, x))
+            x := or(x, shr(2, x))
+            x := or(x, shr(4, x))
+            x := or(x, shr(8, x))
+            x := or(x, shr(16, x))
+            x := or(x, shr(32, x))
+            x := or(x, shr(64, x))
+            x := or(x, shr(128, x))
+            // Take only the highest bit of x
+            x := xor(x, shr(1, x))
```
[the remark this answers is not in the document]
QGarchery: The original idea comes from here: https://graphics.stanford.edu/~seander/bithacks.html#IntegerLogDeBruijn

Less control flow is generally a good idea for efficiency. We've seen it for the other math operations where we removed the `if` statements. I think it computes exactly the same thing otherwise so I don't think it uses another assumption like not needing to set all the bits to 0

test/helpers/DeBruijn.sol
QGarchery: > `key := shr(248, mul(x, seq))`

No shift to the right cannot overflow, and neither shift to the left. Its the multiplication that can (and often will in this cases) overflow. Using assembly prevents the potential oveflow of the multiplication to make the whole call revert

test/TestMath.sol
QGarchery: The function was refactored in #88 to improve testing

src/math/Math.sol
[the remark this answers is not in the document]
MathisGD: >I love it, but can you also manage to make it easily reusable ? I tried passing it as calldata, but it was less efficient idk why.

Indeed, Solidity has not been thought for this and I forget that this was a lib. Forget about this, memory works well here.

test/helpers/DeBruijn.sol
MathisGD: Oh yes sure

MerlinEgalite (approved):
I would add full stops to the end of sentence but lgtm!

src/math/Math.sol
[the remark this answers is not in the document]
MathisGD: Indeed, it works much better.

MathisGD (approved):

Rubilmax (approved):

PR #81, opened 2022-12-27
Add generalized meta tx contract

### LINKED ISSUE #80: Implement meta transactions re-usable contract?

MerlinEgalite:
Fixes #80 

To do/answer:

- [ ] Verify that safety checks are not missing
- [x] Should we make `MetaTransactionUpgradeable` not to inherit from `ContextMixin`?
- [ ] Should `MetaTransactionUpgradeable` have gaps at the end?
- [ ] Add tests

### DIFF
--- a/src/meta-tx/MetaTransactionUpgradeable.sol
+++ b/src/meta-tx/MetaTransactionUpgradeable.sol
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: GNU AGPLv3
+pragma solidity ^0.8.0;
+
+import {MinimalForwarderUpgradeable} from "openzeppelin-contracts-upgradeable/contracts/metatx/MinimalForwarderUpgradeable.sol";
+import {ContextMixin} from "./ContextMixin.sol";
+
+contract MetaTransactionUpgradeable is ContextMixin, MinimalForwarderUpgradeable {
+    error FunctionCallFailed();
+
+    function executeMetaTransaction(ForwardRequest calldata req, bytes calldata signature)
+        public
+        payable
+        returns (bytes memory)
+    {
+        (bool success, bytes memory returnData) = execute(req, signature);
+        if (!success) revert FunctionCallFailed();
+
+        return returnData;
+    }
+}
--- a/src/meta-tx/MetaTransactionUpgradeable.sol
+++ b/src/meta-tx/MetaTransactionUpgradeable.sol
@@ -2,9 +2,8 @@
 pragma solidity ^0.8.0;
 
 import {MinimalForwarderUpgradeable} from "openzeppelin-contracts-upgradeable/contracts/metatx/MinimalForwarderUpgradeable.sol";
-import {ContextMixin} from "./ContextMixin.sol";
 
-contract MetaTransactionUpgradeable is ContextMixin, MinimalForwarderUpgradeable {
+contract MetaTransactionUpgradeable is MinimalForwarderUpgradeable {
     error FunctionCallFailed();
     error OnlySelf();
 
--- a/src/meta-tx/MetaTransactionUpgradeable.sol
+++ b/src/meta-tx/MetaTransactionUpgradeable.sol
@@ -6,12 +6,14 @@ import {ContextMixin} from "./ContextMixin.sol";
 
 contract MetaTransactionUpgradeable is ContextMixin, MinimalForwarderUpgradeable {
     error FunctionCallFailed();
+    error OnlySelf();
 
     function executeMetaTransaction(ForwardRequest calldata req, bytes calldata signature)
         public
         payable
         returns (bytes memory)
     {
+        if (req.to != address(this)) revert OnlySelf();
         (bool success, bytes memory returnData) = execute(req, signature);
         if (!success) revert FunctionCallFailed();
 
--- a/src/meta-tx/MetaTransactionUpgradeable.sol
+++ b/src/meta-tx/MetaTransactionUpgradeable.sol
@@ -1,21 +1,41 @@
 // SPDX-License-Identifier: GNU AGPLv3
 pragma solidity ^0.8.0;
 
+import {ERC2771ContextUpgradeable} from "openzeppelin-contracts-upgradeable/contracts/metatx/ERC2771ContextUpgradeable.sol";
 import {MinimalForwarderUpgradeable} from "openzeppelin-contracts-upgradeable/contracts/metatx/MinimalForwarderUpgradeable.sol";
 
-contract MetaTransactionUpgradeable is MinimalForwarderUpgradeable {
+contract MetaTransactionUpgradeable is MinimalForwarderUpgradeable, ERC2771ContextUpgradeable {
+    /* ERRORS */
+
+    /// @notice Thrown when the `execute` function reverts.
     error FunctionCallFailed();
-    error OnlySelf();
+
+    /* CONSTRUCTOR */
+
+    constructor(address trustedForwarder) ERC2771ContextUpgradeable(trustedForwarder) {}
+
+    /* INITIALIZER */
+
+    function __MetaTransaction_init() internal onlyInitializing {
+        __MinimalForwarder_init();
+    }
+
+    function __MetaTransaction_init_unchained() internal onlyInitializing {}
+
+    /* PUBLIC */
 
     function executeMetaTransaction(ForwardRequest calldata req, bytes calldata signature)
         public
         payable
         returns (bytes memory)
     {
-        if (req.to != address(this)) revert OnlySelf();
         (bool success, bytes memory returnData) = execute(req, signature);
         if (!success) revert FunctionCallFailed();
 
         return returnData;
     }
+
+    /// @dev This empty reserved space is put in place to allow future versions to add new
+    ///      variables without shifting down storage in the inheritance chain.
+    uint256[50] private __gap;
 }
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "lib/aave-v3-core"]
 	path = lib/aave-v3-core
 	url = https://github.com/aave/aave-v3-core
+[submodule "lib/openzeppelin-contracts-upgradeable"]
+	path = lib/openzeppelin-contracts-upgradeable
+	url = https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable
--- a/remappings.txt
+++ b/remappings.txt
@@ -2,3 +2,4 @@
 ds-test/=lib/forge-std/lib/ds-test/src/
 forge-std/=lib/forge-std/src/
 openzeppelin-contracts/=lib/openzeppelin-contracts/
+openzeppelin-contracts-upgradeable/=lib/openzeppelin-contracts-upgradeable/
--- a/src/meta-tx/ContextMixin.sol
+++ b/src/meta-tx/ContextMixin.sol
@@ -4,11 +4,8 @@ pragma solidity ^0.8.0;
 abstract contract ContextMixin {
     function _msgSender() internal view returns (address sender) {
         if (msg.sender == address(this)) {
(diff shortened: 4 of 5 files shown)


MerlinEgalite:
Ready for review to collect feedbacks (I'll add tests and comments)

src/meta-tx/MetaTransactionUpgradeable.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: GNU AGPLv3
+pragma solidity ^0.8.0;
+
+import {MinimalForwarderUpgradeable} from "openzeppelin-contracts-upgradeable/contracts/metatx/MinimalForwarderUpgradeable.sol";
+import {ContextMixin} from "./ContextMixin.sol";
+
+contract MetaTransactionUpgradeable is ContextMixin, MinimalForwarderUpgradeable {
+    error FunctionCallFailed();
```
Rubilmax, line 8:
```suggestion
    error TransactionFailed();
```

Rubilmax (commented):
I'm not sure to understand why the forwarder is inheriting the context mixin, and am a bit confused

If I'm not mistaken, the context mixin's purpose is to be integrated in a meta-tx-ready protocol (such as Morpho), so that the protocol can query which sender the tx is from, and in a safe manner in the case msg.sender is a relayer?

But then why does the context mixin check if `msg.sender == address(this)` because it's not supposed to be used within the context of the relayer, and rather check whether the sender is a trusted relayer or not

Btw, do we have plans to trust specific relayers? Or deploy our own?

src/meta-tx/MetaTransactionUpgradeable.sol
MerlinEgalite: This is wrong IMO function call fail is more accurate since the transaction has not failed until we explicitly revert.

MerlinEgalite:
> I'm not sure to understand why the forwarder is inheriting the context mixin, and am a bit confused
> 
> If I'm not mistaken, the context mixin's purpose is to be integrated in a meta-tx-ready protocol (such as Morpho), so that the protocol can query which sender the tx is from, and in a safe manner in the case msg.sender is a relayer?
> 
> But then why does the context mixin check if `msg.sender == address(this)` because it's not supposed to be used within the context of the relayer, and rather check whether the sender is a trusted relayer or not
> 
> Btw, do we have plans to trust specific relayers? Or deploy our own?

This is one of the question. We can separate both tbh.

The sender must be the `address(this)` as the relayer should have called `executeMetaTransaction` before and then the called is sent from Morpho itself. If `msg.sender != address(this)` then the original sender did not trigger `executeMetaTransaction`. I don't think we'll need trusted railer as in [EIP2771](https://eips.ethereum.org/EIPS/eip-2771) since in our case we want something more general anyone can sign a message to allow another account to do things on its behalf. It's up to the user to do it or not. Is it clearer?

Rubilmax:
If we want to develop an efficient relayer + context according to EIP2771, then we should separate it to follow the EIP

If we don't stick to the workflow drawn in the EIP, then it's a more specific use-case that we should name otherwise, without referencing EIP2771 as it can be confusing (because the call stack would not be the same)

Aside from this, I don't understand the usecase you designed. Is it possible to have a schema of the desired call flow? Including the Morpho protocol, the user, the relayer (if any) and the type of interactions involved between stakeholders (calls/signatures)?

MerlinEgalite:
> If we want to develop an efficient relayer + context according to EIP2771, then we should separate it to follow the EIP

At the start this is not the purpose. The purpose is to implement the account manager on Morpho-AaveV3 as proposed by @pakim249CAL and not the EIP2771. Do you mean that the naming is misleading?

> Aside from this, I don't understand the usecase you designed. Is it possible to have a schema of the desired call flow? Including the Morpho protocol, the user, the relayer (if any) and the type of interactions involved between stakeholders (calls/signatures)?

The flow would be:
- Alice signs a message allowing Bob to supply on her behalf.
- Bob call `executeMetaTransaction` passing the correct arguments AND the message signed.
- The call is checked thanks to the logic in this PR.

Perhaps I'm missing something though, in this case @pakim249CAL might help.

Rubilmax:
Ok so first, I think we are mis-understood because I replaced `forwarder` with `relayer` in my previous comments. This is now fixed.

> Do you mean that the naming is misleading?

I'd instead think it's just that I saw this PR and looked after what work was done on the meta tx subject so I acknowledged we were sticking to standards (which in this case is the EIP)
So in the end it's a lack of communication / I didn't check the correct resource to review this PR

The flow you presented looks like the flow designed in the EIP, where Alice is the sender and Bob is the forwarder
However, the EIP is suggesting:
1. to separate the context verifier & the forwarder
2. to only trust a specific set of forwarders

Here's my analysis:
1. separating allows to save bytecode size ; I don't see what advantage merging provides
2. I don't get why the EIP states that the signer address could be manipulated, because it's verified using the signature (I guess it has something to do with `ecrecover` vulnerabilities?)


What I understand from our discussion is that you'd like to merge the forwarder & the context verifier. What advantage does it provide?
The EIP does allow anyone to be a relay and submit an off-chain signed tx on behalf of someone else, through a forwarder (which in the case of the EIP, should be trusted - but if we want to introduce a trustless forwarder system then I'd say we're jumping into vulnerabilities)

Rubilmax:
> I've separated them.

I disagree: currently, `MetaTransactionUpgradeable` (the forwarder) inherits from `ContextMixin` (the context verifier), just as if we were expecting the integrator to use the context mixin within the forwarder, which is not what the EIP suggests (even beyond the EIP: why would the forwarder need to access the tx signer with the same logic used to access the tx signer at the protocol level?)

To make it clearer:
- `Morpho` should inherit `ContextMixin` (as you stated it), and check for the trusted forwarder (for security purposes, but we need to dig this)
- the forwarder should be completely independent from the protocol, because it has nothing to do with the protocol (it's just a **forwarder**), and thus should not inherit verifiability properties (e.g. the check for the trusted forwarder) ; so I would remove the inheritance because the forwarder should never access the signer the way the protocol (the application) does via `ContextMixin`

MerlinEgalite:
updated bad network

src/meta-tx/MetaTransactionUpgradeable.sol
MathisGD: Agree with Merlin, because only the context can revert without making the whole tx reverting.

Rubilmax:
Here's a summary so far:

- we want to implement a all-in-one context verifier & forwarder

It allows to only have a single address to interact with (Morpho's), enables trust less forwarding (the protocol itself being the forwarder) as well as saving a CALL (compared to the EIP)

However, it includes the forwarding logic into the protocol

I'm not sure if I'm against or in favor of it
What I'm sure of is that if we go for this solution, then I agree `MetaTransactionUpgradeable` should inherit `ContextMixin`

MerlinEgalite:
> Here's a summary so far:
> 
> * we want to implement a all-in-one context verifier & forwarder
> 
> It allows to only have a single address to interact with (Morpho's), enables trust less forwarding (the protocol itself being the forwarder) as well as saving a CALL (compared to the EIP)
> 
> However, it includes the forwarding logic into the protocol
> 
> I'm not sure if I'm against or in favor of it What I'm sure of is that if we go for this solution, then I agree `MetaTransactionUpgradeable` should inherit `ContextMixin`

Ok if I get it right, we must use the EIP and set `trustedForwarder` as `address(this)` in our case. And then the Mixin would check whether `msg.sender` is equal to the `trustedForwarder` or not. Is that correct?

pakim249CAL (dismissed):

pakim249CAL:
I mostly agree with @Rubilmax but I think this solution is also workable. If this is the solution chosen, then the logic in the Morpho contract needs to filter based on msg.sender being the forwarder. But the Morpho contract can also function as its own forwarder with an additional external entry point which I would prefer.

MerlinEgalite:
> I mostly agree with @Rubilmax but I think this solution is also workable. If this is the solution chosen, then the logic in the Morpho contract needs to filter based on msg.sender being the forwarder. But the Morpho contract can also function as its own forwarder with an additional external entry point which I would prefer.

There are 3 solutions (in the context of meta tx):

1. Use the EIP completely and the trusted forwarder is external to Morpho.
2. Use the EIP. Morpho inherits it and the `ContextMixin` (additional call) and the forwarder is set to the Morpho address.
3. Use this current version that Morpho inherits (as well as the `ContextMixin` -- to merge? --) to remove this extra call.

Wdyt?

pakim249CAL:
I'm in favor of solution 3, as it would be redundant in our case to store a forwarder when we know it will be address(this).

MerlinEgalite:
> I'm in favor of solution 3, as it would be redundant in our case to store a forwarder when we know it will be address(this).

Ok and about merging with the ContextMixin? Should we keep them separate? Or should one contract inherits the other?

pakim249CAL:
Since Morpho is the contract that needs to modify msg.sender, it should be the Morpho contract inheriting ContextMixin. The forwarder contract should only be responsible for appending the right msg sender to the calldata.

Rubilmax:
> Since Morpho is the contract that needs to modify msg.sender, it should be the Morpho contract inheriting ContextMixin. The forwarder contract should only be responsible for appending the right msg sender to the calldata.

Disagreed because the ContextMixin in this case is correct only because it's supposed ti be inherited by the forwarder (otherwise the check `msg.sender == address(this)` wouldn't mean anything if the ContextMixin is inherited by anything else than the forwarder)
So these 2 should be merged to me, without distinction

pakim249CAL:
> > Since Morpho is the contract that needs to modify msg.sender, it should be the Morpho contract inheriting ContextMixin. The forwarder contract should only be responsible for appending the right msg sender to the calldata.
> 
> Disagreed because the ContextMixin in this case is correct only because it's supposed ti be inherited by the forwarder (otherwise the check `msg.sender == address(this)` wouldn't mean anything if the ContextMixin is inherited by anything else than the forwarder) So these 2 should be merged to me, without distinction
> 
> Or renamed to ForwarderContextMixin but what's the point of separating them if they are supposed to be inherited by the same contract?

I'm not really sure what you mean. I'm pretty sure the ContextMixin does not need to be used by a forwarder at all. This is how I understand the flow of generalized metatransactions to be:

Let's suppose we have addresses:
A: The Morpho contract
B: The forwarder contract
C: The caller of the contract

The job of the forwarder is to accept a typed struct that has all the parameters needed to call a function, and verify an accompanying signature of C. After signature verification, the forwarder calls the function on the designated target but with address(C) appended to the calldata.

This is just a regular call, so msg.sender when the function is called in contract A is now address(B). At this point, contract A must do additional processing to use a different msg.sender when it is address(B).

This is the only place where ContextMixin comes into play. Contract A is the only contract that needs to extract msg.sender from the calldata. The forwarder contract only has the job of appending the correct msg.sender to the calldata.

The reason I suggest doing it all in the Morpho contract is because there is actually no reason that A and B have to be distinct. Instead of keeping track of address(B) in contract A, contract A can simply use address(this) if it is it's own forwarder.

MerlinEgalite:
This last message seems to me to be the solution 1, am I missing something?

Rubilmax:
I agree the solution you're talking about is the generalized solution shown in the EIP2771 which is also Merlin's solution 1)

I was talking about how we'd implement solution 3), in which the forwarder & the application are merged ; hence why the forwarder should include the context's _msgSender() and not be separated into a separate contract, because it's not supposed to be inherited by anything else than the forwarder (again, under the assumption than we merge the forwarder & the protocol, which is the solution 3 proposed by Merlin)

Benefits are:
- no trusted forwarder (forwarder is the protocol, `address(this)`)
- 1 hot CALL instead of 1 cold CALL

MerlinEgalite:
Re-opening it because I think it can still be cool add it to the repo. I'll clean things if necessary

PR #84, opened 2022-12-27
Use forge fmt & lint-staged

Rubilmax:


README.md
```diff
@@ -2,14 +2,56 @@
...
+    "commands": [
+      {
+        "match": ".sol",
+        "isAsync": true,
+        "cmd": "forge fmt ${file}"
+      }
+    ]
+  }
+}
+```
```
MathisGD, line 51: Not mandatory to have this here, but ok

package.json
```diff
@@ -7,10 +7,15 @@
   "author": "MerlinEgalite <merlin@morpho.xyz>",
   "license": "AGPL-3.0",
   "scripts": {
-    "lint": "prettier --write **.sol"
+    "postinstall": "husky install"
```
MathisGD, line 10: what is this useful for ?

MathisGD (commented):
I think that `forge fmt` works well, but if we don't use it everywhere it might be annoying no ?

README.md
MerlinEgalite: Where do you want to have it?

README.md
```diff
@@ -2,14 +2,56 @@
 
 Repository gathering useful libraries and contracts.
 
-## Setup
+## Getting started
+
+Install this repository as a forge dependency:
+
+```bash
+forge install https://github.com/morpho-dao/morpho-utils
```
MerlinEgalite, line 10:
```suggestion
forge install morpho-dao/morpho-utils
```

It's a bit shorter

MerlinEgalite (approved):
I'm with the changes even though it's a bit annoying that we cannot do the following
```solidity
if (true) return 100;
else return 0;
```

README.md
MathisGD: idk, nowhere ^^
people do whatever they want with their setups

pakim249CAL (approved):

README.md
MerlinEgalite: Ok I think it's more like a tip than something mandatory so I don't it's a pb

MathisGD: yes

package.json
Rubilmax: To install git hooks!

Rubilmax:
> I think that `forge fmt` works well, but if we don't use it everywhere it might be annoying no ?

Why would it be annoying?

MathisGD:
>Why would it be annoying?

Because you need to switch the configuration of your IDE each time you change of repo

Rubilmax:
> Because you need to switch the configuration of your IDE each time you change of repo

Fortunately, VSCode & IntelliJ both provide such features
And open-source contributors (outside the Morpho Labs team) are used to such changes in setup (because all open-source repositories don't use the same formatter/formatting settings)

So I'd not consider this

MathisGD:
You mean that you can set a different formatter for each repo on VSCode ? Without the annoying setting file ?

Rubilmax:
> You mean that you can set a different formatter for each repo on VSCode ? Without the annoying setting file ?

Lol no, via the settings file

MathisGD:
Ok, It's manageable though

Then as you want, but I think that it is easier if we change on all repos

MathisGD (approved):

PR #87, opened 2023-01-02
Replace `iszero(gt(x, constant))` with `lt(x, add(constant, 1))`

### LINKED ISSUE #85: Replace `iszero(gt(x, constant))` with `lt(x, constant + 1)`
https://github.com/morpho-dao/morpho-utils/pull/68#discussion_r1057806970

Rubilmax:
Fixes #85 

Formal proof is failing for unknown reason, being investigated by @QGarchery

### DIFF
--- a/src/math/CompoundMath.sol
+++ b/src/math/CompoundMath.sol
@@ -9,8 +9,8 @@ library CompoundMath {
     /// CONSTANTS ///
 
     // Only direct number constants and references to such constants are supported by inline assembly.
+    uint256 internal constant WAD = 1e18;
     uint256 internal constant MAX_UINT256 = 2 ** 256 - 1;
-    uint256 public constant WAD = 1e18;
 
     /// INTERNAL ///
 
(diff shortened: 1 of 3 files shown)


MathisGD (commented):
I thought that you didn't liked this (operations between constants) 🤔

We might need to increase the number of optimization in order for this optimization to be one

Rubilmax:
> I thought that you didn't liked this (operations between constants) 🤔

Yes I'm not a fan of it, but in certain cases the tradeoff is worth to me (e.g. instead of defining a `uint256` as hex because assembly doesn't support composed constants)
But I'm indeed more open than before

> We might need to increase the number of optimization in order for this optimization to be one

Actually I think it's because I applied the change everywhere whereas sometimes the right operand is not a constant (it depends on the input variable)

MerlinEgalite (approved):

MathisGD:
>Yes I'm not a fan of it, but in certain cases the tradeoff is worth to me (e.g. instead of defining a uint256 as hex because assembly doesn't support composed constants)

It is not the case here, though

Rubilmax:
> It is not the case here, though

Yes it is, you can't save the constant as a composed constant outside assembly?

> The formal verification is failing 🤔

Yes @QGarchery is investigating 😝
> Formal proof is failing for unknown reason, being investigated by @QGarchery

QGarchery (approved):
LGTM.

Concerning the CI fail, I was unable to reproduce it in local so I ran the test again on the CI and it seems to pass now. I'll ask about this and report back here

MathisGD (approved):

PR #88, opened 2023-01-02
Refactoring log2 and its testing

QGarchery:
Gas report seems off when you use the internal functions which should really be private.

See https://github.com/morpho-dao/morpho-utils/pull/89 that shows a 645 gas usage for the new log2 (vs 624 before)

### DIFF
--- a/src/math/Math.sol
+++ b/src/math/Math.sol
@@ -54,7 +54,8 @@ library Math {
     /// A particular de Bruijn sequence.
     uint256 internal constant DE_BRUIJN_SEQ = 0x00818283848586878898a8b8c8d8e8f929395969799a9b9d9e9faaeb6bedeeff;
 
-    /// @dev Returns log2(x) given that x is a power of 2.
+    /// @dev Returns log2(x) given that x is a power of 2 and returns 0 on input 0.
+    /// @dev This computation makes use of De Bruijn sequences, their usage for computing log2 is referenced here: http://supertech.csail.mit.edu/papers/debruijn.pdf
     function _lookupDeBruijn(uint256 x) internal pure returns (uint256 y) {
         assembly {
             // Hash table associating the first 256 powers of 2 to their log
@@ -77,9 +78,7 @@ library Math {
     }
 
     /// @dev Returns the floor of log2(x) and returns 0 on input 0.
-    /// @dev This computation makes use of De Bruijn sequences, their usage for computing log2 is referenced here: http://supertech.csail.mit.edu/papers/debruijn.pdf
-    function log2(uint256 x) internal pure returns (uint256 y) {
-        x = _roundDownToPowerOf2(x);
-        y = _lookupDeBruijn(x);
+    function log2(uint256 x) internal pure returns (uint256) {
+        return _lookupDeBruijn(_roundDownToPowerOf2(x));
     }
 }
--- a/src/math/Math.sol
+++ b/src/math/Math.sol
@@ -52,7 +52,7 @@ library Math {
     }
 
     /// A particular de Bruijn sequence.
-    uint256 internal constant deBruijnSeq = 0x00818283848586878898a8b8c8d8e8f929395969799a9b9d9e9faaeb6bedeeff;
+    uint256 internal constant DE_BRUIJN_SEQ = 0x00818283848586878898a8b8c8d8e8f929395969799a9b9d9e9faaeb6bedeeff;
 
     /// @dev Returns log2(x) given that x is a power of 2.
     function _lookupDeBruijn(uint256 x) internal pure returns (uint256 y) {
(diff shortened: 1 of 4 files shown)


MathisGD (commented):
I'm not a big fan tbh (always the same debate). I must admit that it simplifies the tests quite a bit, but the fact that we add a useless return in the code for this reason looks also very weird.

QGarchery:
> I'm not a big fan tbh (always the same debate). I must admit that it simplifies the tests quite a bit, but the fact that we add a useless return in the code for this reason looks also very weird.

Agreed, it looks weird. Adding the return actually saves some gas somehow

src/math/Math.sol
```diff
@@ -48,7 +47,14 @@ library Math {
             x := or(x, shr(128, x))
             // Take only the highest bit of x
             x := xor(x, shr(1, x))
+        }
+        return x;
+    }
 
+    /// @dev Returns a particular de Bruijn sequence and log2(x) given that x is a power of 2.
+    function _lookupDeBruijn(uint256 x) internal pure returns (uint256 deBruijnSeq, uint256 y) {
+        deBruijnSeq = 0x00818283848586878898a8b8c8d8e8f929395969799a9b9d9e9faaeb6bedeeff;
```
MerlinEgalite, line 56: Why is it returned?

MerlinEgalite (commented):
It seems that the changes are motivated by tests, is it correct? 
What is the impact on gas? I'm not sure to understand your comment below. On this PR I'm seeing 734 as gas cost for the `log2`.

> Gas report seems off when you use the internal functions which should really be private.
> See https://github.com/morpho-dao/morpho-utils/pull/89 that shows a 645 gas usage for the new log2 (vs 624 before)

src/math/Math.sol
MathisGD: I think it is for testing purpose..

Rubilmax: We could have it set as a constant & use it in a test, right?

MathisGD: Oh yes good point ! 😆

QGarchery:
> It seems that the changes are motivated by tests, is it correct?
What is the impact on gas? I'm not sure to understand your comment below. On this PR I'm seeing 734 as gas cost for the log2.

If you look at the PR linked you see that the impact on gas is 645 gas, whereas before it was 624. The linked PR does not change the library, but only how it is used. Even then, nothing is changed if we run only the `CompareLog2` tests. So I'm assuming the compilation is different because the compiler sees that the functions `_roundDownToPowerOf2` and `_lookupDeBruijn` are used and cannot be inlined. This is why I said they those 2 functions should really be private (but then we couldn't test them)

src/math/Math.sol
QGarchery: Ok, done ! Gas usage is now 658

src/math/Math.sol
```diff
@@ -63,9 +71,15 @@ library Math {
...
             let key := shr(248, mul(x, deBruijnSeq)) // With the multiplication it works also for 0
             y := shr(248, mload(add(m, key))) // Look in the table and take the first byte
         }
     }
+
+    /// @dev Returns the floor of log2(x) and returns 0 on input 0.
+    /// @dev This computation makes use of De Bruijn sequences, their usage for computing log2 is referenced here: http://supertech.csail.mit.edu/papers/debruijn.pdf
+    function log2(uint256 x) internal pure returns (uint256 y) {
+        x = _roundDownToPowerOf2(x);
+        y = _lookupDeBruijn(x);
```
MathisGD, line 83: Does this save any gas ?
```suggestion
        y = _lookupDeBruijn(_roundDownToPowerOf2(x));
```

src/math/Math.sol
```diff
@@ -33,9 +33,8 @@ library Math {
         }
     }
 
-    /// @dev Returns the floor of log2(x) and returns 0 on input 0.
-    /// @dev This computation makes use of De Bruijn sequences, their usage for computing log2 is referenced here: http://supertech.csail.mit.edu/papers/debruijn.pdf
-    function log2(uint256 x) internal pure returns (uint256 y) {
+    /// @dev Returns the biggest power of 2 smaller than or equal to x and returns 0 on input 0.
+    function _roundDownToPowerOf2(uint256 x) internal pure returns (uint256) {
```
Rubilmax, line 37:
```suggestion
    function _log2Down(uint256 x) internal pure returns (uint256) {
```

(this is a suggestion to make it less verbose)

QGarchery: I don't really like it because it seems like we are applying log2 and then rounding down (but instead we find the power of 2 that is immediatly less than the input)

src/math/Math.sol
QGarchery: Yes a little bit, it's changed now

src/math/Math.sol
MathisGD: Agree with @QGarchery. Otherwise `_onlyHighestSetBit` works too, but not especially better.

Rubilmax (approved):

QGarchery:
Seen with @MerlinEgalite: we are closing this for now as we feel it makes it less efficient while not providing enough clarity.

PR #103, opened 2023-06-02
Add `abs`

Rubilmax:


MerlinEgalite (approved):
Nice! what's the gas imrpovement?

pakim249CAL:
What does safeAbs protect from?

pakim249CAL (approved):

Rubilmax:
> What does safeAbs protect from?

It protects from reverting in the case we've reached the negative bound (which doesn't have an exact positive opposite)

Rubilmax:
> Nice! what's the gas imrpovement?

It is less than 10 gas lol

MerlinEgalite:
> > Nice! what's the gas imrpovement?
> 
> It is less than 10 gas lol

lol ok thx

Tristan22400 (approved):

src/math/Math.sol
```diff
@@ -6,6 +6,24 @@ pragma solidity ^0.8.0;
 /// @custom:contact security@morpho.xyz
 /// @dev Library to perform simple math manipulations.
 library Math {
+    function abs(int256 x) internal pure returns (int256 y) {
```
makcandrov, line 9: why not returning an `uint256`?

makcandrov (approved):

src/math/Math.sol
makcandrov: but returning a `uint256` instead of an `int256` has several advantages:
- the function will always be exact and will never revert, even in the `type(int256).min` edge case.
- if the user needs a `uint256`, returning a `uint256` directly eliminates the conversion step, it improves efficiency since the default conversion from signed to unsigned integers performs a check on the sign, which is unnecessary in this case.
- if the user needs an `int256`, returning a `uint256` allows them to freely handle the edge case. They have the flexibility to perform the conversion at no additional cost if they know that the resulting amount will fit within the `int256` range. Alternatively, they can apply special logic in the edge case.

Also, `abs(type(int256).min) = 2 ** 255` which only fits in 256 bits.

QGarchery: I agree that the function would be very clean if it returned an uint256, but I think that the assumption @Rubilmax made is sound and that there is a use case for staying in `int256`. If the user wanted to use `uint256` instead, then there are other functions to do that. For example the subtraction could be using the `zeroFloorSub` (from the current file).

Rubilmax: But now it seems both functions have equivalent advantages 😁 so I'd like to implement your suggestion: shall I name it `uabs` ?

QGarchery: Maybe `unsignedAbs` ?

QGarchery (approved):

PR #105, opened 2023-06-07
Add fuzz finding revert on percentmul

QGarchery:


Rubilmax:
I fixed the failing tests because test ranges were too wide. In particular, when rounding half up, it is necessary that the test value be less than `type(uint256).max - half_scale`, which wasn't necessarily the case.

pakim249CAL (approved):

makcandrov (approved):

Rubilmax:
I shouldn't have merged this. Let's do this optimistically: if you've got changes requests, let's discuss them here first and I'll open a dedicated PR after.

PR #107, opened 2023-06-07
CVL2 migration

QGarchery:


Rubilmax (commented):
I infer that CVL2 requires not defining functions in specs?
What's the goal of this PR? What issue is it trying to solve pls?

QGarchery:
> What's the goal of this PR? What issue is it trying to solve pls?

The migration to CVL2 is today. We can still use CVL1 but it's getting deprecated, and so it will be difficult to maintain in the future. I'm starting to work on the translation of our specs to CVL2 for that reason, and also because there is one bug fix that we need to prove the properties of the `abs` functions (it has to do with the negation of mathematical integers)

> I infer that CVL2 requires not defining functions in specs?

You can define functions in specs in CVL2 but there is an issue with the definition of `MIN_INT256`. It seems to come from a check that prevents creating signed integers that are too big or too small, I'm still investigating

QGarchery:
Closing this in favor of https://github.com/morpho-org/morpho-utils/pull/109, where we don't try to add new tests

PR #108, opened 2023-06-08
Initial deposit to custom recipient

Rubilmax:
- fixes https://github.com/spearbit-audits/review-morpho-june/issues/25

### DIFF
--- a/src/ERC4626UpgradeableSafe.sol
+++ b/src/ERC4626UpgradeableSafe.sol
@@ -21,6 +21,7 @@ abstract contract ERC4626UpgradeableSafe is ERC4626Upgradeable {
 
     /* INITIALIZER */
 
+    /// @dev Initializes the ERC4626 vault with the given name, symbol and decimals and mints `initialDeposit` shares to the given recipient.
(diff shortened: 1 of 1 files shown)


Tristan22400 (dismissed):

pakim249CAL (dismissed):

makcandrov (approved):

makcandrov (dismissed):

MerlinEgalite (approved):
Maybe we should specify NATSPECs for end users

Tristan22400 (approved):

QGarchery (approved):
I can imagine that the deployer of the vault cares about precision (and so wants to make an initial deposit) but does not have the skim function. In that case it would be ok to mint the shares to the vault, but now it's not always practical to know the address of the vault that is being created. This is probably only a small issue though

Rubilmax:
> but now it's not always practical to know the address of the vault that is being created. This is probably only a small issue though

I don't get it: in this case, just pass `address(this)` at initialization?

makcandrov (approved):

pakim249CAL (approved):

PR #109, opened 2023-06-08
Migration to CVL2 (second attempt)

QGarchery:


certora/specs/compoundMath.spec
```diff
@@ -1,34 +1,35 @@
 methods {
-    mul(uint256, uint256)   returns (uint256)   envfree
-    div(uint256, uint256)   returns (uint256)   envfree
+    function mul(uint256, uint256)   external returns (uint256)   envfree;
+    function div(uint256, uint256)   external returns (uint256)   envfree;
 }
 
 definition WAD()  returns uint = 10^18;
+definition UINT_LIMIT() returns mathint = 2 ^ 255 * 2;
```
MathisGD, line 7: what about
```suggestion
definition MAX() returns mathint = 2 ^ 255 * 2;
```

certora/specs/compoundMath.spec
```diff
@@ -1,34 +1,35 @@
...
 
 /// mul ///
 
 rule mulSafety(uint256 x, uint256 y) {
     uint res = mul(x, y);
 
-    assert res == x * y / WAD(); 
+    assert to_mathint(res) == x * y / WAD();
```
MathisGD, line 14: I understand that know math integers and "solidity integers" are not comparable as is, you need to specify the type conversion is that correct ? And what happens if you try the opposite conversion (math => solidity) ?

MathisGD (approved):

certora/specs/compoundMath.spec
Rubilmax: Why can't we `2 ^ 256`? 😆

certora/specs/compoundMath.spec
QGarchery: >And what happens if you try the opposite conversion (math => solidity) ?

I don't think this functionality exists, it would be weird because mathints are obtained from solidity integers

certora/specs/compoundMath.spec
Rubilmax: Sorryy

certora/specs/compoundMath.spec
MerlinEgalite: `to_mathint` is an internal function in certora?

certora/specs/compoundMath.spec
QGarchery: I find `MAX` to be a bit misleading because the max of a set should be in that set

MathisGD: ah yes indeed. We could use 2**256 - 1 but that may not be practical for tests after

MerlinEgalite (approved):

certora/specs/compoundMath.spec
QGarchery: We will switch back to the previous notation anyway once the issue is resolved on the Certora side

certora/specs/wadRayMath.spec
```diff
@@ -208,39 +209,39 @@ rule rayToWadLiveness(uint256 a) {
...
 
-    assert res == a * WADTORAY();
+    assert to_mathint(res) == a * WADTORAY();
 }
 
 rule wadToRayLiveness(uint256 a) {
     wadToRay@withrevert(a);
 
-    assert lastReverted <=> a * WADTORAY() > 2^256;
+    assert lastReverted <=> a * WADTORAY() > UINT_LIMIT();
```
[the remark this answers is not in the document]
QGarchery: Both are equivalent (because `WADTORAY()` is not a factor of `UINT_LIMIT()`) but I agree that it would be clearer as you suggested. I'll open a PR on top to not lose the approvals

MathisGD (approved):

PR #111, opened 2023-06-12
Refactor constant comments

QGarchery:
Fixes:
- https://github.com/morpho-org/morpho-utils/pull/109#discussion_r1226241637
- https://github.com/morpho-org/morpho-utils/pull/104#discussion_r1225826989
  by moving the comment up, it should apply to the whole constant section

### DIFF
--- a/src/math/CompoundMath.sol
+++ b/src/math/CompoundMath.sol
@@ -7,8 +7,8 @@ pragma solidity ^0.8.0;
 /// @dev Library to perform Compound's multiplication and division in an efficient way.
 library CompoundMath {
     /* CONSTANTS */
-
     // Only direct number constants and references to such constants are supported by inline assembly.
+
     uint256 internal constant WAD = 1e18;
     uint256 internal constant MAX_UINT256 = 2 ** 256 - 1;
 
--- a/certora/specs/wadRayMath.spec
+++ b/certora/specs/wadRayMath.spec
@@ -215,7 +215,7 @@ rule wadToRaySafety(uint256 a) {
 rule wadToRayLiveness(uint256 a) {
     wadToRay@withrevert(a);
 
-    assert lastReverted <=> a * WADTORAY() > UINT_LIMIT();
+    assert lastReverted <=> a * WADTORAY() >= UINT_LIMIT();
 }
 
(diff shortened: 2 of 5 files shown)


MathisGD (dismissed):

src/math/CompoundMath.sol
```diff
@@ -7,8 +7,8 @@ pragma solidity ^0.8.0;
 /// @dev Library to perform Compound's multiplication and division in an efficient way.
 library CompoundMath {
     /* CONSTANTS */
-
     // Only direct number constants and references to such constants are supported by inline assembly.
+
```
MerlinEgalite, line 11: Did you try, with an auto generated doc? I'm not certain it will work

Rubilmax (approved):

src/math/CompoundMath.sol
QGarchery: I tried and those comments do not show, but I think we do not want to show them, because the user should not care about how the implementation uses inline assembly.
That being said, I don't know how to have a comment that shows for a section, and not for a particular function/variable. Do you know how to do that @MerlinEgalite ?

Thinking about it, I'm not sure why the math constants are internal and not private in all our libraries. It's weird that the constant `MAX_UINT256` is defined both in `CompoundMath` and `PercentageMath`, no ? Wdyt @morpho-org/onchain-team ?

MerlinEgalite: because sometimes you would like to access the scaling factor such as `WAD`.

> It's weird that the constant MAX_UINT256 is defined both in CompoundMath and PercentageMath

What is you suggestion?

QGarchery: > What is you suggestion?

I'll create a PR on top to show it.

What's your thoughts on the other questions @MerlinEgalite: showing or not showing the comment, how to show a comment for a whole section

QGarchery: #112

pakim249CAL (approved):

makcandrov (approved):

MathisGD (approved):
`/* CONSTANTS */` and `/* INTERNAL */` don't seem mandatory such simple libraries but ok to have them. LGTM

PR #112, opened 2023-06-13
Factorize math constants

QGarchery:
The purpose of this PR is to avoid defining the same constants multiple times, so they are all defined at the same place and used when needed

### DIFF
--- a/src/math/CompoundMath.sol
+++ b/src/math/CompoundMath.sol
@@ -1,19 +1,13 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity ^0.8.0;
 
+import "./MathConstants.sol";
+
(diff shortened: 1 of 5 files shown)


Rubilmax (dismissed):

makcandrov (approved):

pakim249CAL (approved):

Tristan22400 (approved):

QGarchery:
I don't have a strong opinion on it, but could you explain what you don't like about it @MathisGD @MerlinEgalite ?

pakim249CAL (approved):

QGarchery:
Closing as it's not completely necessary, and it may be worth it to not factorize simple libraries

PR #114, opened 2023-06-20
Add `EIP712`, `Ownable` & `Ownable2Step`

Rubilmax:


src/ERC2330.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import {IERC2330} from "src/interfaces/IERC2330.sol";
+
+/// @dev Gas-optimized extsload getters to allow anyone to read storage from this contract.
+///      Enables the benefit of https://eips.ethereum.org/EIPS/eip-2330 without requiring changes to the execution layer.
+contract ERC2330 is IERC2330 {
```
pakim249CAL, line 8: Would it be appropriate to set these contracts as abstract?

src/access/Ownable2Step.sol
```diff
@@ -0,0 +1,48 @@
...
+    /* PUBLIC */
+
+    /// @inheritdoc IOwnable2Step
+    function pendingOwner() public view virtual returns (address) {
+        return _pendingOwner;
+    }
+
+    /// @inheritdoc IOwnable2Step
+    function acceptOwnership() public virtual {
+        address sender = msg.sender;
```
pakim249CAL, line 25: No need to cache the sender because msg.sender is actually cheaper than using memory.

